Issue #14, opened 2022-08-18
Emoji data for Scribe apps


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Now that [Scribe-iOS](https://github.com/scribe-org/Scribe-iOS) is adding autosuggest in [#194](https://github.com/scribe-org/Scribe-iOS/issues/194) and autocomplete in [#188](https://github.com/scribe-org/Scribe-iOS/issues/188), other additions for these features could be considered including emojis. This feature would be to create python scripts that would create arrays of words and emojis that can represent them. When one of these words is entered, the emojis could then serve as autosuggestions or completions (depending on if the user has pressed space or not respectively).

These scripts would make use of [emoji](https://github.com/carpedm20/emoji/) for the representations of emojis as words in the given languages, and would also make use of [nltk](https://github.com/nltk/nltk) for stemming (not lemmatizing) words to derive `smile` from `smiling`. Other language packages could be leveraged to derive adjective forms like `smiley`, and convert words like types of trees (aspen, birch, etc) to "tree" so they can also be converted (likely another issue).

### Contribution

I'd be happy to work with someone on this once the features in Scribe-iOS are finished :)

The corresponding Scribe-iOS issue for this is [#51](https://github.com/scribe-org/Scribe-iOS/issues/51), which would need to be worked on in the foreseeable future for this issue to make sense.

andrewtavis:
Note that final version of this would be included in [update_data.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/load/update_data.py) so that changes to the scripts would be reflected in the apps.

andrewtavis:
Adding some easter eggs into this feature could also be a nice touch. These might be best as just autosuggestions and not corrections. Some ideas are:

- Wikidata: 💚 and 📁
- Wikipedia: ❤️ and 📚
- WIkimedia: 🌐 and 📖
- Scribe: 💙 and ✏️

wkyoshida:
Hey @andrewtavis - I'd love to help out with this!

[...]

PR #22, opened 2022-11-03
#14 setup process unicode

wkyoshida:
Hey @andrewtavis!

Added some quick setup for the emoji work.

**Some notes:**
- While it should be possible to get the _'emoji-to-keyword'_ relationships from Unicode, i.e.
```json
      "😀": {
        "default": [
          "face",
          "grin",
          "grinning face"
        ],
        "tts": [
          "grinning face"
        ]
      },
```
(example from [here](https://github.com/unicode-org/cldr-json/blob/f27f55f1dd487af7cf4260f56296ee1c7649b7fc/cldr-json/cldr-annotations-full/annotations/en/annotations.json#L3283-L3292)),

I'm potentially thinking that the inverse, _'keyword-to-emoji',_ would be useful, i.e. resulting in something of the like:  
```
# below not json, only for illustration
{
  face: [😀, ...],
  grin: [😀, ...],
  grinning: [😀, ...],
}
```
This might be useful perhaps, since the trigger for the autosuggestion would be the keyword, not the emoji. References off of the keyword could be more important. 
- Also, I'm thinking that storing the Unicode Code Point for the emoji, as opposed to one of its rendered image varieties, might also make more sense. For the emoji in the above point, I believe it would be `U+1F600`. This is more to perhaps facilitate the use of the emoji data independent of the platform or OS - Android and iOS use different images for emojis, but the Unicode code is the same. The code could be provided and the platform renders the emoji as it would.
- Haven't added anything `pyicu` yet, as still looking into if it could be easily used for this work.

Let me know any thoughts on the above points :rocket:

### DIFF
--- a/src/scribe_data/extract_transform/process_unicode.py
+++ b/src/scribe_data/extract_transform/process_unicode.py
@@ -1,0 +1,47 @@
+"""
+Process Unicode
+------------
+
+Module for processing Unicode based corpuses for autosuggestion generation.
+
+Contents:
+    gen_emoji_autosuggestions
+"""
+
+def gen_emoji_autosuggestions(
+    language="English",
+    num_emojis=500,
+    ignore_keywords=None,
+    update_scribe_apps=False,
+    verbose=True,
+):
+    """
+    Generates a dictionary of keywords (keys) and emoji unicodes(s) associated with them (values).
+
+    Parameters
+    ----------
+        language : string (default=en)
+            The language autosuggestions are being generated for.
+
+        num_emojis: int (default=500)
+            The number of emojis that autosuggestions should be generated from.
+
+        ignore_keywords : str or list (default=None)
+            Keywords that should be ignored.
+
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Makes sense to have this file 😊 Thanks for putting it together, @wkyoshida!

andrewtavis:
> I'm potentially thinking that the inverse, _'keyword-to-emoji',_ would be useful

Makes total sense, @wkyoshida 😊 Yes, we'll have to do a little bit of work to "invert" the relationships, but it shouldn't be too hard :)

> * Also, I'm thinking that storing the Unicode Code Point for the emoji, as opposed to one of its rendered image varieties, might also make more sense. For the emoji in the above point, I believe it would be `U+1F600`. This is more to perhaps facilitate the use of the emoji data independent of the platform or OS

Also 100% agree with the above, and I appreciate that you always think in terms of platform agnosticism 🚀

I'll have some time to look at this tomorrow evening!

Issue #26, opened 2022-11-17
Discussion: future of Scribe-Data as a package


wkyoshida:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Opening this issue to continue the discussion from [here](https://github.com/scribe-org/Scribe-Data/pull/25#issuecomment-1315742719) and [here](https://github.com/scribe-org/Scribe-Data/pull/25#issuecomment-1316038439). I thought it could be good to move it to a separate location where it could continue even after the linked PR gets merged.

I'm thinking this issue can be used to generally discuss what Scribe-Data could provide to a greater community as a package. Some related topics could include:
- What about converting Scribe-Data as a language pack-generation tool available as a package?
- What about simply making the language packs themselves as the packages instead?
- Where to make the packages available? Currently in pypi
- Possibly renaming the packages from `scribe-data` to something more descriptive of their purpose?
- What changes would be needed to make Scribe-Data less tightly coupled to the other Scribe repos?
- Whatever becomes of Scribe-Data, how to integrate it into the rest of Scribe to continue serving Scribe's data needs? 

Some thoughts were already made in the linked PR, but we can resume here. I agree that Scribe-Data could potentially become something useful for a larger audience :people_holding_hands: :raised_hands: Continuing this discussion could be good to see what that could be.

andrewtavis:
Hi @wkyoshida :) Firstly, apologies for my relative absence the last couple days. Lots going on between getting healthy, work and the rest of life... 😊

This issue will definitely help, thanks for it! I had discussions enabled on Scribe repos before, but I feel that it only divides the discussion until we actually have more conversations happening. As far as the next two points are concerned:

> * What about converting Scribe-Data as a language pack-generation tool available as a package?
> * What about simply making the language packs themselves as the packages instead?

Would we be able to do both? Say that we have up to date versions of the packs that are available for direct download, and the package also has Python codes that generate the packs? We could keep the current structure of `extract_transform` and `load` in this case?

> * Whatever becomes of Scribe-Data, how to integrate it into the rest of Scribe to continue serving Scribe's data needs?

To me it makes sense that we make downloading the packs general and adapt the way that language packs are downloaded within the app. Should be doable 😊

[...]

PR #27, opened 2022-11-23
#14 emoji popularity rank

wkyoshida:
**Description:**
- Added `2021_ranked.tsv` as a resource file. 
- Added processing to include the emojis' popularity rank.
  - Emojis not in the ranking/resource file, are just `None` for now 
- Added minor formatting edits to comments, i.e. periods.

**Notes:**
- `setup.py` changes
  - The new line `package_data={"": ["2021_ranked.tsv"]},` is to save the resource file as part of the package.  
  - Upped the minimum python version to 3.9 since that's the version the following used standard library is first available: 
    - `from importlib.resources import files`
    - Is there any concern with this? Just checking if there is a reason to actually keep Scribe-Data at python 3.6 
    - Also updated the versions in the `classifiers`
- Next steps:
  - Add the step the save the generated files in the other Scribe repos 
    - I'm thinking this could potentially change with whatever comes out of #26, but that's something for another day.
    - This implementation is likely fine for now as that's what is done for the other data generation.


andrewtavis:
Wonderful! Will check this soon and then will start implementing the iOS functionality 😊

Thanks so much! 🚀

andrewtavis:
> * Is there any concern with this? Just checking if there is a reason to actually keep Scribe-Data at python 3.6

There's no reason for it :) I've been accustomed to just doing 3.6 for things as that was the standard when I was first doing packages two years ago. Just haven't updated it since 😊

> * The new line `package_data={"": ["2021_ranked.tsv"]},` is to save the resource file as part of the package.

Nice! Very interesting :)

> * This implementation is likely fine for now as that's what is done for the other data generation.

Totally fine for now. We have lots to discuss in #26, and from there we can figure out what we want to have happen :)

andrewtavis (approved):
Looks and works great, @wkyoshida! Thanks for all the efforts 😊 Really looking forward to such a significant v2.2.0!

Issue #61, opened 2024-01-27
Fix MacOS build due to failing dependency installations


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

As of now the build for all versions of MacOS is failing. This was originally because of Tensorflow, which has now been removed as a dependency in favor of other packages that were also being used and had some similar functionality (parallel downloads). The most recent error is:

```
ERROR: Could not build wheels for marisa-trie, which is required to install pyproject.toml-based projects
Successfully built PyICU sentencepiece
Failed to build marisa-trie
```

Would be great if we could figure this out!

andrewtavis:
CC @wkyoshida, in case you have some ideas on this :)

shashank-iitbhu:
The workflow is failing because one of the dependancies `language-data` depends on `marisa-trie<0.8.0 and >=0.7.7`.
And `marisa-trie<0.8.0 and >=0.7.7` does not have support for `Python 3.12` which is the error in the `CI logs`.
As mentioned in the [marisa-trie docs](https://pypi.org/project/marisa-trie/), `marisa-trie 1.1.0` has added support for python 3.12 .

The solution is to remove the `language-data` dependancy from `requirements.txt` as I think it is not being used anywhere in the project.

I also tried pinning both `language-data` and `marisa-trie` to their latest versions but this won't work beacuse of this:

```ERROR: Cannot install language-data==1.1 and marisa-trie>=1.1.0 because these package versions have conflicting dependencies.
The conflict is caused by:
    The user requested marisa-trie>=1.1.0
    language-data 1.1 depends on marisa-trie<0.8.0 and >=0.7.7

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts```

shashank-iitbhu:
@andrewtavis @wkyoshida

andrewtavis:
[...]

PR #63, opened 2024-02-12
Update utils.py

Jk40git:
Change the _find's docstring to _load_json's

### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/src/scribe_data/utils.py
+++ b/src/scribe_data/utils.py
@@ -5,6 +5,8 @@ Update Utils
 Utility functions for data extraction, formatting and loading.
 
 Contents:
+    _load_json,
+    _find,
     get_scribe_languages,
     get_language_qid,
     get_language_iso,
@@ -76,7 +78,7 @@ _languages = _load_json(
 
 
 def _find(source_key: str, source_value: str, target_key: str, error_msg: str):
-       """
+    """
     Each 'language', (english, german,..., etc) is a dictionary of key/value pairs:
 
         entry = {
@@ -87,16 +89,19 @@ def _find(source_key: str, source_value: str, target_key: str, error_msg: str):
             "ignore-words": [...]
         }
 
-    Given a key/value pair, the 'source', and the 'target' key, get the 'target' value.
+    Given a key/value pair, the 'source' and the 'target' key get the 'target' value.
 
     Parameters
     ----------
         source_value : str
             The source value to find equivalents for (e.g. 'english').
+
         source_key : str
             The source key to reference (e.g. 'language').
+
         target_key : str
             The key to target (e.g. 'iso').
+
         error_msg : str
             The message displayed when a value cannot be found.
 
--- a/src/scribe_data/utils.py
+++ b/src/scribe_data/utils.py
@@ -76,30 +76,22 @@ _languages = _load_json(
 
 
 def _find(source_key: str, source_value: str, target_key: str, error_msg: str):
-    """
-    Each 'language', (english, german,..., etc) is a dictionary of key/value pairs:
+    """Loads a JSON resource from a package into a python entity.
 
-        entry = {
-            "language": "english",
-            "iso": "en",
-            "qid": "Q1860",
-            "remove-words": [...],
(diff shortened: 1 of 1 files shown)


andrewtavis:
I see there was some confusion here, @Jk40git. Big thing is that we don't want the docstring to "be the docstring of the other one", but we want them to have the same style. For find we currently have:

```
    """
    Each 'language', (english, german,..., etc) is a dictionary of key/value pairs:

        entry = {
            "language": "english",
            "iso": "en",
            "qid": "Q1860",
            "remove-words": [...],
            "ignore-words": [...]
        }

    Given a key/value pair, the 'source', and the 'target' key, get the 'target' value.

    Args:
        source_value (str): e.g. 'english'.
        source_key (str): e.g. 'language'.
        target_key (str): e.g. 'iso'.
        error_msg (str): for when a value cannot be found.

    Raises:
        ValueError: when a source_value is not supported.

    Returns:
        The 'target' value given the passed arguments.
    """
```

What this should be is:

```
    """
    Each 'language', (english, german,..., etc) is a dictionary of key/value pairs:

        entry = {
            "language": "english",
            "iso": "en",
            "qid": "Q1860",
            "remove-words": [...],
            "ignore-words": [...]
        }

    Given a key/value pair, the 'source', and the 'target' key, get the 'target' value.

    Parameters
    -----------
        source_value : str
            The source value to find equivalents for (e.g. 'english').
        source_key : str
            The source key to reference (e.g. 'language').
        target_key : str
            The key to target (e.g. 'iso').
        error_msg : str
            The message displayed when a value cannot be found.

    Raises
    -------
        ValueError : when a source_value is not supported.

    Returns
    --------
        The 'target' value given the passed arguments.
    """
```

The big thing is that everywhere that we have `Args:` in a docstring we want:

```
Parameters
-----------
    {LIST_OF_THE_ARGS}
```

And everywhere we have `Returns:` we want:

```
Returns
--------
    {LIST_OF_RETURN_VALUES}
```

Can you convert the docstring of `_find` over to what I have above, and then I can send along a list of functions that we also need to do other conversions for?

Jk40git:
Just made an edit [#63 ](https://github.com/Jk40git/Scribe-Data-1/blob/patch-1/src/scribe_data/utils.py)

andrewtavis:
Nice, yes, this is looking good @Jk40git! Here are the other docstrings that need edits:

- In [checkquery.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/checkquery.py) basically all for function docstrings need to be edited
- In [utils.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/utils.py) we need `_find` as you already did and then `check_command_line_args` and `check_and_return_command_line_args`

Want to send along another commit with all of these changes, and then we can close this up? In cases where there's a section that's not `Args:` or `Returns:`, let's just keep it in order and put dashes under it. So for example `Raises:` we can also do:

```
Raises
------
```

as in the example above :) Let me know if there are any questions on this!

Jk40git:
> Nice, yes, this is looking good @Jk40git! Here are the other docstrings that need edits:
> 
> * In [checkquery.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/checkquery.py) basically all for function docstrings need to be edited
> * In [utils.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/utils.py) we need `_find` as you already did and then `check_command_line_args` and `check_and_return_command_line_args`
> 
> Want to send along another commit with all of these changes, and then we can close this up? In cases where there's a section that's not `Args:` or `Returns:`, let's just keep it in order and put dashes under it. So for example `Raises:` we can also do:
> 
> ```
> Raises
> ------
> ```
> 
> as in the example above :) Let me know if there are any questions on this!

Do you want me to change the  `check_command_line_args` and `check_and_return_command_line_args` with the same docstring from _find?

andrewtavis:
Don't worry about the Mac builds, by the way. This is known in #61 :)

Issue #67, opened 2024-02-18
Add reference to the architecture diagrams


wkyoshida:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue is to add a blurb in the `README.md` informing there are architecture diagrams (in the Organization's [`ARCHITECTURE.md`](https://github.com/scribe-org/Organization/blob/main/ARCHITECTURE.md)) available with some overviews of the Scribe projects.

NOTE:
- The goal with the blurb is primarily to provide some helpful initial guidance to newcomers.
- This same blurb can later be copied then into other project's `README.md`s and elsewhere.

Jk40git:
Working on this

Jk40git:
@andrewtavis  I have just created a pull request for this issue please check and give me a feedback

wkyoshida:
Closing via https://github.com/scribe-org/Scribe-Data/pull/82

Thanks @Jk40git! :rocket:

andrewtavis:
Thanks for the initial commit @Jk40git and for the edits, @wkyoshida! I'll send it along to the other repos :)

PR #82, opened 2024-02-27
Added  reference to the architecture diagrams  in Readme.md

Jk40git:
### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER
  #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@
 
 Scribe applications are available on [iOS](https://github.com/scribe-org/Scribe-iOS), [Android](https://github.com/scribe-org/Scribe-Android) (WIP) and [Desktop](https://github.com/scribe-org/Scribe-Desktop) (planned).
 
+# Architecture Diagrams 
+
+Here is an overview of the [architecture diagrams](https://github.com/scribe-org/Organization/blob/main/ARCHITECTURE.md) of our organization, the services, the 
+applications, the processes and the external systems and sources with which it interacts.
+
+  It depicts the projects [Scribe-org](https://github.com/scribe-org) is developing and the relationship between the 
+  projects and external systems and sources.  
+  
 <a id="contents"></a>
 
 # **Contents**
@@ -32,7 +40,6 @@ Scribe applications are available on [iOS](https://github.com/scribe-org/Scribe-
 - [Environment Setup](#environment-setup)
 - [Supported Languages](#supported-languages)
 - [Featured By](#featured-by)
-- [Architecture Diagrams](#architecture-diagrams)
 
 <a id="process"></a>
 
@@ -231,9 +238,3 @@ Many thanks to all the [Scribe-Data contributors](https://github.com/scribe-org/
   <br>
 </div>
 
-<a id="architecture-diagrams"></a>
-# Architecture Diagrams  [`⇧`](#contents)
-Scribe[[1]](https://en.wikipedia.org/wiki/Scribe) is a community of language enthusiasts, learners and developers creating tools to help people communicate with confidence.We strive to be an inclusive community, welcoming coders and open-source supporters of any level.  
- Here is the overview of the [architecture diagrams](https://github.com/scribe-org/Organization/blob/main/ARCHITECTURE.md#full-architecture) of our organization, the services, the applications, the processes and the external systems and sources with which it interacts. 
-  It depicts the projects [Scribe-org](https://github.com/scribe-org) is developing and the relationship between the 
-  projects and external systems and sources.  
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Scribe applications are available on [iOS](https://github.com/scribe-org/Scribe-
 - [Environment Setup](#environment-setup)
 - [Supported Languages](#supported-languages)
 - [Featured By](#featured-by)
+- [Architecture Diagrams](#architecture-diagrams)
 
 <a id="process"></a>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -229,3 +230,10 @@ Many thanks to all the [Scribe-Data contributors](https://github.com/scribe-org/
   &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
   <br>
 </div>
+
+<a id="architecture-diagrams"></a>
+# Architecture Diagrams  [`⇧`](#contents)
+Scribe[[1]](https://en.wikipedia.org/wiki/Scribe) is a community of language enthusiasts, learners and developers creating tools to help people communicate with confidence.We strive to be an inclusive community, welcoming coders and open-source supporters of any level.  
+ Here is the overview of the [architecture diagrams](https://github.com/scribe-org/Organization/blob/main/ARCHITECTURE.md#full-architecture) of our organization, the services, the applications, the processes and the external systems and sources with which it interacts. 
+  It depicts the projects [Scribe-org](https://github.com/scribe-org) is developing and the relationship between the 
+  projects and external systems and sources.
```
wkyoshida, line 239: Hey! :wave: 
Sorry - perhaps, I might have specified more in the issue description.

I was actually thinking the blurb about the architecture diagrams could just be at the very end of the introductory section of the `README.md`, i.e. right after the 

> Scribe applications are available on iOS...

That made sense to me, since the introduction section is already pointing out the other Scribe projects. Right after that, having the blurb about the architecture diagrams showing how the projects then connect kinda made sense logically to me. 

Two notes on the above:
- That would also eliminate the need to have to restate in the blurb what Scribe is as well.
- The other `README.md`s also follow a similar format for their introductory sections, which is why the above made sense to me also.

README.md
```diff
@@ -229,3 +230,10 @@ Many thanks to all the [Scribe-Data contributors](https://github.com/scribe-org/
   &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
   <br>
 </div>
+
+<a id="architecture-diagrams"></a>
+# Architecture Diagrams  [`⇧`](#contents)
+Scribe[[1]](https://en.wikipedia.org/wiki/Scribe) is a community of language enthusiasts, learners and developers creating tools to help people communicate with confidence.We strive to be an inclusive community, welcoming coders and open-source supporters of any level.  
+ Here is the overview of the [architecture diagrams](https://github.com/scribe-org/Organization/blob/main/ARCHITECTURE.md#full-architecture) of our organization, the services, the applications, the processes and the external systems and sources with which it interacts.
```
wkyoshida, line 237: When linking to the architecture diagrams, I'm thinking that it makes sense to simply link to the full file anyways.
So the `#full-architecture` in the url can get omitted here.

README.md
```diff
@@ -229,3 +230,10 @@ Many thanks to all the [Scribe-Data contributors](https://github.com/scribe-org/
   &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
   <br>
 </div>
+
+<a id="architecture-diagrams"></a>
+# Architecture Diagrams  [`⇧`](#contents)
```
wkyoshida, line 235: The table of contents references can get omitted as well, since the idea would be to simply add the blurb within the introductory section, at the end.

wkyoshida (commented):
Some thoughts for this PR @Jk40git, which I realize might have made more sense if I had specified before in the issue... Apologies for that!

Anyways, thank you for taking this on!

README.md
Jk40git: @wkyoshida Noted with thanks

README.md
```diff
@@ -23,6 +23,14 @@
 
 Scribe applications are available on [iOS](https://github.com/scribe-org/Scribe-iOS), [Android](https://github.com/scribe-org/Scribe-Android) (WIP) and [Desktop](https://github.com/scribe-org/Scribe-Desktop) (planned).
 
+# Architecture Diagrams
```
wkyoshida, line 26: We can simply remove this here too as that would create a header for a new section.
The idea is to just make the below blurb part of the introductory section.

README.md
```diff
@@ -23,6 +23,14 @@
 
 Scribe applications are available on [iOS](https://github.com/scribe-org/Scribe-iOS), [Android](https://github.com/scribe-org/Scribe-Android) (WIP) and [Desktop](https://github.com/scribe-org/Scribe-Desktop) (planned).
 
+# Architecture Diagrams 
+
+Here is an overview of the [architecture diagrams](https://github.com/scribe-org/Organization/blob/main/ARCHITECTURE.md) of our organization, the services, the 
+applications, the processes and the external systems and sources with which it interacts.
+
+  It depicts the projects [Scribe-org](https://github.com/scribe-org) is developing and the relationship between the 
+  projects and external systems and sources.
```
wkyoshida, line 32: I'd also remove all the newlines here as well and just put all the text in a single line - to keep with the formatting of the rest of the `README.md`. This text is all the same paragraph anyways.

wkyoshida (commented):
Thanks for staying on top of the changes @Jk40git !

wkyoshida:
Thanks for this @Jk40git!
I'll make a couple minor tweaks after merging, but then we should be good to add this blurb in the other repos as well
@andrewtavis we could simply do that then via your batch-edits

wkyoshida (approved):

Issue #69, opened 2024-02-24
Extract noun gender functions to utils.py


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As a part of refactoring the codebase, it would be great if the [map_genders function](https://github.com/scribe-org/Scribe-Data/blob/851410d678c61536735900c7b85df95dbafe5a58/src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py#L36) that's used in various noun formatting processes was extracted to [utils.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/utils.py) and reused. This would be done for all currently finished noun formatting processes that have genders, so French, German, Italian, Portuguese, Russian, Spanish and Swedish.

### Contribution

Happy to discuss the work to be done and support implementation!

ikeadeoyin:
Hello @andrewtavis,
I am interested in working on this.

andrewtavis:
Assigned, and thanks for all the effort and collaboration so far!

andrewtavis:
Closed via #84 🚀 Thanks so much for the contribution, @ikeadeoyin!

PR #84, opened 2024-02-27
feat: add map gender function to utils.py

ikeadeoyin:
### Contributor checklist

---

### Description

Added the map_gender function to `utils.py`

### Related issue

- #69

### DIFF
--- a/src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
+++ b/src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
@@ -9,6 +9,7 @@ import collections
 import json
 import os
 import sys
+
 PATH_TO_SCRIBE_ORG = os.path.dirname(sys.path[0]).split("Scribe-Data")[0]
 PATH_TO_SCRIBE_DATA_SRC = f"{PATH_TO_SCRIBE_ORG}Scribe-Data/src"
 sys.path.insert(0, PATH_TO_SCRIBE_DATA_SRC)
--- a/src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
+++ b/src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
@@ -9,6 +9,11 @@ import collections
 import json
 import os
 import sys
+PATH_TO_SCRIBE_ORG = os.path.dirname(sys.path[0]).split("Scribe-Data")[0]
+PATH_TO_SCRIBE_DATA_SRC = f"{PATH_TO_SCRIBE_ORG}Scribe-Data/src"
+sys.path.insert(0, PATH_TO_SCRIBE_DATA_SRC)
+
+from scribe_data.utils import map_genders
 
 LANGUAGE = "German"
 QUERIED_DATA_TYPE = "nouns"
@@ -31,25 +36,6 @@ else:
 
 with open(data_path, encoding="utf-8") as f:
     nouns_list = json.load(f)
-
-
-def map_genders(wikidata_gender):
-    """
-    Maps those genders from Wikidata to succinct versions.
-
-    Parameters
-    ----------
-        wikidata_gender : str
-            The gender of the noun that was queried from WikiData
-    """
-    if wikidata_gender in ["masculine", "Q499327"]:
-        return "M"
-    elif wikidata_gender in ["feminine", "Q1775415"]:
-        return "F"
-    elif wikidata_gender in ["neuter", "Q1775461"]:
-        return "N"
-    else:
-        return ""  # nouns could have a gender that is not valid as an attribute
 
 
 def order_annotations(annotation):
--- a/src/scribe_data/utils.py
+++ b/src/scribe_data/utils.py
@@ -17,7 +17,8 @@ Contents:
     get_android_data_path,
     get_desktop_data_path,
     check_command_line_args,
-    check_and_return_command_line_args
+    check_and_return_command_line_args,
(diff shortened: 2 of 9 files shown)


src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
```diff
@@ -9,6 +9,11 @@
 import json
 import os
 import sys
+PATH_TO_SCRIBE_ORG = os.path.dirname(sys.path[0]).split("Scribe-Data")[0]
+PATH_TO_SCRIBE_DATA_SRC = f"{PATH_TO_SCRIBE_ORG}Scribe-Data/src"
+sys.path.insert(0, PATH_TO_SCRIBE_DATA_SRC)
```
wkyoshida, line 14: Sorry - I guess I'm a little confused as to why the above is needed. Does something not work without it?

src/scribe_data/utils.py
```diff
@@ -420,3 +420,21 @@ def check_and_return_command_line_args(
         python {all_args[0]} '["comma_separated_sets_in_quotes"]'
         """
     )
+
+def map_genders(wikidata_gender):
```
wkyoshida, line 424: If I'm understanding https://github.com/scribe-org/Scribe-Data/issues/69 correctly, it seems that what we'd rather want here actually is that `map_genders()` also works for all supported languages. There are other `map_genders()` functions for other languages that exist elsewhere in the code, as pointed out in the issue. The changes here only address the `map_genders()` for German.

So to make the function general and reconcile all supported languages, it could make sense to modify `map_genders(wikidata_gender)` into something like `map_genders(wikidata_gender, iso_code)` instead perhaps - where `iso_code` is `"en"`, `"fr"`, `"es"`, etc. We can then use the `iso_code` for the language-specific checks, such as `"Q499327"` here  for German masculine.

Let me know if there are any questions regarding the above :smile:

src/scribe_data/utils.py
```diff
@@ -420,3 +420,21 @@ def check_and_return_command_line_args(
...
+            The gender of the noun that was queried from WikiData
+    """
+    if wikidata_gender in ["masculine", "Q499327"]:
+        return "M"
+    elif wikidata_gender in ["feminine", "Q1775415"]:
+        return "F"
+    elif wikidata_gender in ["neuter", "Q1775461"]:
+        return "N"
+    else:
+        return ""  # nouns could have a gender that is not valid as an attribute
```
wkyoshida, line 440: nit:
Add a newline at the end of file

---

There are some uncommon issues that can occasionally occur with the absence of an ending newline 🙃

wkyoshida (commented):
Thanks so much for taking this on @ikeadeoyin! :rocket: 

Added some comments to the PR! :v:

src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
ikeadeoyin: I am unable to import the `map_genders` function in the `utils.py` file without these lines. It gives an error message `(ModuleNotFound: no module named scribe_data)`

src/scribe_data/utils.py
ikeadeoyin: Thank you so much for the review, it was helpful in understanding the codebase. I can see that we have `map_genders()` function for other languages.

I have tried implementing this feedback however on running `update.py` file, I got a couple of files modified and a new file created(`src/scribe_data/extract_transform/languages/German/nouns/nouns_queried.json`).

Can I commit these extra files?

src/scribe_data/utils.py
ikeadeoyin: I will do this, thanks.

src/scribe_data/utils.py
andrewtavis: Let's not commit the extra files like `nouns_queried.json`, @ikeadeoyin :) These should have been deleted by the `format_nouns.py` file in the German directory. The way it works is that the query results are brought down, they're formatted and saved in the final version, and then the original file is deleted. Let's just have this focus on the Python files 😊

src/scribe_data/extract_transform/languages/German/nouns/format_nouns.py
andrewtavis: I'll work to figure this out after the merge, @ikeadeoyin :)

src/scribe_data/utils.py
ikeadeoyin: Alright @andrewtavis,

Following the contribution guide, I have committed the Python files however, I have this error
```
Scribe-Data git:(add-map-gender-function) ✗ git pull --rebase upstream main
error: cannot pull with rebase: You have unstaged changes.
error: Please commit or stash them. 
```

wkyoshida: Hey @ikeadeoyin :wave: 
Are the unstaged changes the `*.json` files?
If so, I guess you could either delete them from your local machine or run `git stash` to [stash](https://www.atlassian.com/git/tutorials/saving-changes/git-stash#stashing-your-work) them so that your working directory is clean before pulling

andrewtavis (approved):
We still have the weird sys import magic at the top of the files, but this will be worked out when I got through and convert all of the formatting processes over to being lexeme based rather than string based 😊 Thanks for all the work here, @ikeadeoyin! Great to be improving the packages structure like this 🚀

Issue #57, opened 2023-11-14
Add readthedocs documentation to Scribe-Data


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to add in [readthedocs](https://about.readthedocs.com/?ref=readthedocs.com) for Scribe-Data. Goal would be to have all functions and classes automatically documented. 

- Note that this issue is linked to #56 😊

### Contribution

Happy to support with this! We should discuss implementation beforehand, but I've used readthedocs extensively for other projects and am happy to help with setup or add parts to the docs.

shashank-iitbhu:
assign @andrewtavis

andrewtavis:
Thank you @shashank-iitbhu!

m-charlton:
Apologies for the late reply. Started to look at this late last year but, other things got in the way.
Will gather my notes and lay out a plan of action

andrewtavis:
Great to hear from you, @m-charlton! Plan is to get the baseline for this up for @shashank-iitbhu later today. We can talk about this here in the issue and also check into the work for the doc strings in #56 :)

andrewtavis:
Ok all 😊 We now have the base docs available at [scribe-data.readthedocs.io](https://scribe-data.readthedocs.io/), and the GitHub integration is working fine on commits to the repo 🙌 In setting it up, for some reason the modules weren't being loaded in though. I'm wondering if maybe we didn't have the current `__init__.py` structure when the last release was made? Not sure. My assumption is that the following in [utils.rst](https://github.com/scribe-org/Scribe-Data/blob/main/docs/source/utils.rst) should work:

```
utils
=====

The :py:mod:`utils` module provides utility functions for data extraction, formatting and loading.

.. automodule:: scribe_data.utils
    :members:
    :private-members:
```

What we're getting though is:

```
WARNING: autodoc: failed to import module 'utils' from module 'scribe_data'; the following exception was raised:
No module named 'scribe_data.utils'
```

[...]

PR #99, opened 2024-03-10
Fix for sphinx autodoc

shashank-iitbhu:
### Contributor checklist

---

### Description

- Fix for sphinx autodoc not finding scribe_data module.
- modified `sys.path` in `conf.py`

### Related issue

- #57

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -18,7 +18,7 @@ import sys
 #
 import sphinx_rtd_theme
 
-sys.path.insert(0, os.path.abspath(".."))
+sys.path.insert(0, os.path.abspath("../../src"))
 
 # -- Project information -----------------------------------------------------
 


shashank-iitbhu:
Getting all the documentation for checkquery and utils.
Screenshot 2024-03-10 at 3 59 28 PM

docs/source/conf.py
```diff
@@ -18,7 +18,7 @@
 #
 import sphinx_rtd_theme
 
-sys.path.insert(0, os.path.abspath(".."))
+sys.path.insert(0, os.path.abspath("../../src"))
```
wkyoshida, line 21: :raised_hands: Nice!!

Quick question - trying to catch up on our progress with the docs.. just wondering about the `extract_transform` and `load` stuff. Do those get addressed with this change as well or is the plan for us to tackle them later?

shashank-iitbhu: Can apply the same for `extract_transform` also, `extract_wiki.py` and `process_wiki.py` have some helper functions too.
not sure about `load`.

andrewtavis (approved):
Thanks for this, @shashank-iitbhu! It was working for me as well before this, but then looking at it this does seem to be the right path from conf.py :) Will test some things out with this merged in!

Issue #109, opened 2024-03-18
Repository Python code formatting PR check


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

A great addition to the pull request workflow would be to add a check for code formatting to make sure that all code is formatted with [Ruff](https://github.com/astral-sh/ruff). We'd want a file in `.github/workflows` that's called `pr_ci.yaml`, with the formatting check generally able to be based off of the check from [another project I work on](https://github.com/activist-org/activist/blob/main/.github/workflows/pr_ci_backend.yaml).

### Contribution

Would be happy to support with this! A contributor has already asked to work on it :)

andrewtavis:
Hey @Linfye, this is the issue I was saying I can make for you :) If it's still interesting, can you write in here so I can assign? Thanks, and let me know if you have any questions!

Linfye:
@andrewtavis I have submitted a [PR](https://github.com/scribe-org/Scribe-Data/pull/111) for this issue. And I'm not very proficient in GitHub actions and Ruff, so please let me know if there are any problems. 😊

andrewtavis:
Nice, @Linfye! I'll check out the PR and make the fixes to the current code :) :) We also need to update the maintainer workflow so that we also make sure that these checks pass.

andrewtavis:
Closed by #111 🥳 Thank you, @Linfye! 😊

PR #111, opened 2024-03-19
Add GitHub Actions workflow for backend code checks

Linfye:
### Contributor checklist

---

### Description

I add GitHub Actions workflow for code formatting checks with Ruff.

### Related issue

- #ISSUE_NUMBER
#109


andrewtavis (approved):
This is working really really well, @Linfye! Thanks for the efforts here! I added in the line `# noqa: E402` in many places to suppress an error in that we're usually running Scribe-Data from the command line, so this is why the normal src package structure isn't working. Things to consider later 🤔

Great to have this up!

Issue #107, opened 2024-03-18
Add Scribe favicon to the Scribe-Data docs


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

One thing that would be great to add to the docs is that they could have a custom `.ico` favicon. [This Stack Overflow question](https://stackoverflow.com/questions/54639192/change-the-favicon-of-the-sphinx-read-the-docs-theme) gives a general rundown of how we could achieve this.

### Contribution

Happy to send along a `.ico` or `.png` file for this and someone can implement! Note that this image should be stored in the `source/_static` directory in the `docs` directory.

mrbazzan:
Hi @andrewtavis, I would love to work on this

andrewtavis:
Nice @mrbazzan! Thanks for your interest in helping out :) :) I'll send along a `.ico` file for this later on today and then you can take it from there?

mrbazzan:
Sure, no problem.
I'll be waiting for the `.ico` file then

spalominor:
> Nice @mrbazzan! Thanks for your interest in helping out :) :) I'll send along a `.ico` file for this later on today and then you can take it from there?

Hey, I want to contribute to your project, can we talk about that? Tell what to do.

andrewtavis:
Thanks for your interest, @spalominor! This issue is taken, but there should be some new good first issues being made soon. Keep an eye on the issues for new ones, and I'll try to remember to let you know when something comes that you could work on. You can also join us in the community Matrix chats, with the description of how to do that being in the readme 😊

spalominor:
> Thanks for your interest, @spalominor! This issue is taken, but there should be some new good first issues being made soon. Keep an eye on the issues for new ones, and I'll try to remember to let you know when something comes that you could work on. You can also join us in the community Matrix chats, with the description of how to do that being in the readme 😊

Perfect @andrewtavis, I will see for new issues to contribute.

andrewtavis:
[...]

PR #113, opened 2024-03-19
Add favicon to the docs

mrbazzan:
### Contributor checklist

---

### Description

This is how it looks like:

Screenshot 2024-03-20 at 00 06 09

### Related issue

- #107


docs/source/conf.py
```diff
@@ -188,6 +188,9 @@
     "display_version": True,
 }
 
+# Adding favicon to the docs
```
andrewtavis, line 191: Very minor nitpick for me that you don't need to worry about, @mrbazzan :) For comments that are their own line we try to have them always be a complete sentence with a period at the end. If it's an inline comment then we don't capitalize the first word or include a period. These are just the habits that I picked up from looking into Python coding standards :)

I'll fix this!

andrewtavis (approved):
Really nice, @mrbazzan! Appreciate you following up for the icon and getting to this as quickly as you did 😊

Issue #115, opened 2024-03-20
Edit and expand Arabic data processes


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would check and expand the queries and related data processes found in the [scribe_data/extract_transform/languages/Arabic](https://github.com/scribe-org/Scribe-Data/tree/main/src/scribe_data/extract_transform/languages/Arabic) directory. It would be great to start with expanding the queries found in the nouns and verbs directory, and then from there we can discuss a formatting process :) The [query_nouns.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/extract_transform/languages/Arabic/nouns/query_nouns.sparql) and [query_verbs.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/extract_transform/languages/Arabic/nouns/query_nouns.sparql) files can be expanded on using the similar queries found for other languages. The formatting process should wait until the formatting process is expanded to focus on individual lexemes.

### Contribution

Happy to support someone who has interest in working on this! 😊

andrewtavis:
CC @mrbazzan who had interest in this! Please write in the issue and I'll assign! Also let me know if there are any questions :)

mrbazzan:
I would like to start with this one.

andrewtavis:
Fantastic, @mrbazzan! Let me know if there's anything we can do to help :)

mrbazzan:
All I can think of right now is to copy some of the optional queries
from [German's query_nouns.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/extract_transform/languages/German/nouns/query_nouns.sparql) to Arabic's own but I don't
really understand what that does or what's going on.

Is there anything else to be considered?

andrewtavis:
Look into some of the Arabic lexemes and see if there are other properties or statements that could be added. There might not be though, so feel free to send along the basic files converted from German to Arabic!

mrbazzan:
> Look into some of the Arabic lexemes and see if there are other properties or statements that could be added. 

What do you mean? I tried a couple of queries and they contain a lot of duplicate data.

[...]

PR #127, opened 2024-04-03
Expand verb query for Arabic data

mrbazzan:

Based on Wikidata, Arabic Verbs are divided into three (which are further divided into):
- `Past`: Performative Passive and Performative Active
- `Present (Fi'il mudari')`: Indicative Present, Subjunctive Present, Jussive Present
- `Imperative`

These are just my assumptions though.

I added queries for a tense in each category. Am I on the right track?

By the way, is it possible to add a feature to the 
`Wikidata Query Service` so that some part of a query can be collapsed?
This is so that there isn't constant scrolling between the editor and the result
tab below (if some of the queries are collapsed, everything can appear on the
screen at the same time).

- #115

### DIFF
--- a/src/scribe_data/extract_transform/languages/Arabic/verbs/query_verbs_1.sparql
+++ b/src/scribe_data/extract_transform/languages/Arabic/verbs/query_verbs_1.sparql
@@ -7,7 +7,9 @@ SELECT
   ?verb
   ?presFPS ?presSPSM ?presSPSF ?presTPSM ?presTPSF
   ?presSPD ?presTPDM ?presTPDF
-  ?presFPP ?presSPPM ?presSPPF WHERE {
+  ?presFPP ?presSPPM ?presSPPF
+
+WHERE {
   ?lexeme a ontolex:LexicalEntry ;
     dct:language wd:Q13955 ;
     wikibase:lexicalCategory wd:Q24905 ;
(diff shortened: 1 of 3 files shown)


andrewtavis:
> I added queries for a tense in each category. Am I on the right track?

Is looking good to me, @mrbazzan!

> By the way, is it possible to add a feature to the
> Wikidata Query Service so that some part of a query can be collapsed?
> This is so that there isn't constant scrolling between the editor and the result
> tab below (if some of the queries are collapsed, everything can appear on the
> screen at the same time).

The Wikidata Query Service UI doesn't offer this feature, which is why I normally am always writing my SPARQL queries in VS Code. There are a couple of good extensions that you can install to make it all go a bit easier, which you can see at the top of the [environment setup section of the readme](https://github.com/scribe-org/Scribe-Data?tab=readme-ov-file#environment-setup-).

andrewtavis (approved):
Brining this in regardless of the fail as there are no changes to the environment and the error is a network timeout. Thanks for this, @mrbazzan! Looking forward to the next queries! We'd definitely welcome work like this for nouns, verbs, etc for any language you'd feel like sending along :)

Let us know if you need further assistance with this 😊

mrbazzan:
> The Wikidata Query Service UI doesn't offer this feature, which is why I normally am always writing my SPARQL queries in VS Code. 

:(

> There are a couple of good extensions that you can install to make it all go a bit easier, which you can see at the top of the [environment setup section of the readme](https://github.com/scribe-org/Scribe-Data?tab=readme-ov-file#environment-setup-).

Thanks. The `QID` extension looks nice.

Issue #147, opened 2024-06-07
Implement CLI `--total` functionality


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would implement the `--total` (`-t`) functionality of the Scribe-Data CLI. This functionality would check Wikidata for the total of certain groupings of languages and word types. Usage of this would be:

```bash
scribe-data total -l German -wt nouns  # number of German noun lexemes
scribe-data total -l German  # number of German lexemes
scribe-data total -wt nouns  # number of noun lexemes
```

- Note: it would be good to allow the user to pass `nouns` or `noun`, etc, in order to avoid unneeded errors :)

The following Python code could be edited for most of the functionality that we need for this, whereby we could also add some changes such that the `word_type` argument would also function :) From there the result of this function is returned to the user with a message including the given language and/or word types.

```py
from SPARQLWrapper import SPARQLWrapper, JSON

def get_total_lexemes(language, word_type):
    endpoint_url = "https://query.wikidata.org/sparql"
    sparql = SPARQLWrapper(endpoint_url)

    # SPARQL query template.
    query_langage_template = """
    SELECT 
        (COUNT(DISTINCT ?lexeme) as ?total)

    WHERE {{
      VALUES ?language { wd:{} }
      ?lexeme dct:language ?language ;
              wikibase:lexicalCategory ?category .
    """

    filter_word_template  = """
         FILTER(?category IN ( { wd:{} } ))
    """

    end_of_query = """
    }}
    """

    if word_type:
        query_langage_template += filter_word_template

    query_langage_template += end_of_query

    # Replace {} in the query template with the language value.
    query = query_langage_template.format(language)  # , word_type  # <-- we want to include this and have this also be repalcesd

    sparql.setQuery(query)
    sparql.setReturnFormat(JSON)
    results = sparql.query().convert()

    return int(results["results"]["bindings"][0]["total"]["value"])
```

- Note: the above function also needs to be able to accept lists, so it should be `languages` and `word_types` :)

### Contribution

@mhmohona will be working on this as a part of GSoC 2024! ☀️ Please write in here so I can assign, and let us know if there's anything we can do to support!

[...]

PR #162, opened 2024-06-26
Implement CLI `--total` functionality

mhmohona:
### Contributor checklist

---

### Description

Implemented the `--total (-t)` functionality would check Wikidata for the total of certain groupings of languages and word types.

image

### Related issue

Fixes - #147


mhmohona:
It still has a problem. It is recognizing data-type as language like this - 
image

andrewtavis:
Any thoughts on what's causing it, @mhmohona? :)

src/scribe_data/cli/total.py
```diff
@@ -19,3 +19,93 @@
...
+    "german": "Q188",
+    "italian": "Q652",
+    "portuguese": "Q5146",
+    "russian": "Q7737",
+    "spanish": "Q1321",
+    "swedish": "Q9027"
+}
+
+# Dictionary to map data types to their Wikidata Q-IDs
+data_type_to_qid = {
```
mhmohona, line 38: @andrewtavis will you please check here if I have the correct QID?

src/scribe_data/cli/total.py
```diff
@@ -19,3 +19,93 @@
...
+        The input string representing a language or data type.
+
+    Returns
+    -------
+    str or None
+        The QID corresponding to the input string, or None if not found.
+    """
+    if input_str:
+        input_str_lower = input_str.lower()
+        if input_str_lower in language_to_qid:
```
andrewtavis, line 61: Putting a space before all `else` statements and maybe between the `elif` and the `return` would be good :)

andrewtavis:
Minor comments so far:
- Let's include the language and the data type in the output
- From @wkyoshida, something like:

```bash
Language: German
Data type: Verbs
Total number of lexemes: 999
```

- Let's move `data_type_to_qid` to the `data_type_metadata.json` file
- Let's reference the `language_metadata.json` file for the `language_to_qid` information

We'll be good to go after all this! 🥳

andrewtavis:
Quick check here @mhmohona, are you planning on getting to the changes we mentioned above?

wkyoshida:
> It still has a problem. It is recognizing data-type as language like this - image

Doesn't this work since `-l German` is passed first? Could just be that the first argument is recognized and used. The second may have been passed, but it's disregarded. 

Looks like there is a [choices](https://docs.python.org/3/library/argparse.html#choices) parameter that we can pass in to `add_argument()` to specify what are the allowable options for an argument. Maybe we look if specifying this could make sense as the mechanism for controlling valid/invalid inputs?

mhmohona:
Currently stuck in this comment 😢 
> Let's move data_type_to_qid to the data_type_metadata.json file

andrewtavis:
By this I mean let's move the functionality of `data_type_to_qid` to the `data_type_metadata.json` file such that we just import the data at the top of the file rather than using a custom function :)

mhmohona:
> > It still has a problem. It is recognizing data-type as language like this - image
> 
> Doesn't this work since `-l German` is passed first? Could just be that the first argument is recognized and used. The second may have been passed, but it's disregarded.
> 
> Looks like there is a [choices](https://docs.python.org/3/library/argparse.html#choices) parameter that we can pass in to `add_argument()` to specify what are the allowable options for an argument. Maybe we look if specifying this could make sense as the mechanism for controlling valid/invalid inputs?

@wkyoshida, thank you for looking into it. I have solved this problem.

mhmohona:
> By this I mean let's move the functionality of `data_type_to_qid` to the `data_type_metadata.json` file such that we just import the data at the top of the file rather than using a custom function :)

I need help with the QID. :( Unable  to find the correct QIDs. It would be super helpful for me if you would update the `data_type_metadata.json` file with QIDs @andrewtavis.

src/scribe_data/cli/main.py
```diff
@@ -127,7 +128,11 @@ def main() -> None:
     )
     total_parser._actions[0].help = "Show this help message and exit."
     total_parser.add_argument(
-        "-lang", "--language", type=str, help="The language(s) to check totals for."
+        "-l",
```
andrewtavis, line 131: I'm going to remove `-l` as an option, @mhmohona, as I think that making sure that we don't have commands and arguments that are the same is important :)

src/scribe_data/cli/main.py
```diff
@@ -190,7 +195,13 @@ def main() -> None:
         )
 
     elif args.command in ["total", "t"]:
-        return
+        if not args.language and not args.data_type:
```
andrewtavis, line 198: Really appreciate the quality of your messages to the user, @mhmohona!

src/scribe_data/cli/main.py
```diff
@@ -190,7 +195,13 @@ def main() -> None:
...
+            print(
+                "Error: At least one of -lang/--language or -dt/--data-type must be specified."
+            )
+            total_parser.print_help()
+            return
```
andrewtavis, line 203: Let's always do a space after `return` :)

andrewtavis (approved):
Thanks so much for this, @mhmohona! The full experience of Scribe-Data is really coming along 😊

Issue #155, opened 2024-06-09
Update update_data.py tqdm progress bar only on a successful stage


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now the [update_data.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/wikidata/update_data.py) process allows will use a try-except to run the query [here](https://github.com/scribe-org/Scribe-Data/blob/1981f8210739429ddf515ec259a2de5083a90aae/src/scribe_data/wikidata/update_data.py#L150):

```py
try:
    results = sparql.query().convert()
except HTTPError as err:
    print(f"HTTPError with {query_path}: {err}")
```

If the except block is entered, we still update the progress bar, but actually the exception triggers a retry later in the process. Ideally the progress bar would not be updated at this stage.

### Contribution

Would be great to get some help here! Happy to support or get to it myself eventually 😊

henrikth93:
I will take a look at it!

andrewtavis:
Thanks, @henrikth93! As discussed, would be good to even comment out the query functionality for a moment while you work on this and then figure out how to only update the progress bar on a "successful query" :)

axif0:
@andrewtavis, hello.. can I give a try on this?

andrewtavis:
Hey @axif0 👋 This one is already assigned to @henrikth93. If he's fine with you taking over, then we can switch the assignment, or we can also try to find you a new issue :)

henrikth93:
Hello @axif0 and @andrewtavis! I am looking into this. If we can not find a new issue for @axif0, then maybe we can look into it together?

andrewtavis:
@axif0, do you want to check out #96? For this what would likely be best would be to get the articles from Wikidata for the given language and then filter them out from the result 😊 Would be happy to discuss this further!

axif0:
Thanks @andrewtavis, I have commented there. Will you please have a look at the comment section?

andrewtavis:
Closed by #176 :) Thanks, @henrikth93!

PR #176, opened 2024-07-19
Changed tqdm loop so that we are using "with"

henrikth93:
Changed the tqdm loop so that it is used together with "with", it enables us to prevent printing the progress bar on a failure/error.

### Contributor checklist

---

### Related issue

- #155


andrewtavis:
Quick note on GitHub, @henrikth93: It's best to request review in these cases rather than assign. I'm not really sure what assign would be in this case, but we don't really use it as a convention. Maybe in the future when we have more than one person working on a PR 😊

andrewtavis (approved):
Checked this locally and it's working great, @henrikth93! Thanks for taking this on! 🚀

PR #181, opened 2024-07-28
Added type hints

to-sta:
### Contributor checklist

---

### Description
Last time, I saw there were some missing type hints in the `utils.py` file. I just added the rest of them and used `Any` if it was not clear what will be returned 😄.

### Related issue

- #No related issue

### DIFF
--- a/src/scribe_data/cli/cli_utils.py
+++ b/src/scribe_data/cli/cli_utils.py
@@ -22,7 +22,7 @@ Utility functions for the Scribe-Data CLI.
 
 import json
 from pathlib import Path
-from typing import Dict, List, Union
+from typing import Union
 
 LANGUAGE_METADATA_FILE = (
     Path(__file__).parent.parent / "resources" / "language_metadata.json"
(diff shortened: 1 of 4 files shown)


to-sta:
Starting from Python 3.9, built-in collection types can be used directly as generics. Since Python 3.9 is used I convert to the built-in `list` type.

andrewtavis:
@to-sta, 0c87380 was valid to do given your comment [here](https://github.com/scribe-org/Scribe-Data/pull/181#issuecomment-2254477248), correct? I tool out all type imports of `List` and `Dict` and just used the built-ins instead :)

to-sta:
Yeah, that is correct 👌🏻. Here are the [release notes](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections), in case you want to read more about.

andrewtavis (approved):
Nice, thanks for the confirmation and for some further great help to the project, @to-sta! 🚀 Great to have your influence here 😊

PR #183, opened 2024-08-11
Remove all hardcoded final directories and minor docs updates

andrewtavis:
### Contributor checklist

---

### Description

Sending along a PR so we can see the tests are passing. What I'm doing here is changing the paths so they're referencing centrally saved default directories so that if they do need to change, that we can change them in one place :)

### Related issue

- No related issue


andrewtavis:
CC @mhmohona, just so you have an overview of these changes :)

andrewtavis:
Tests passed, so bringing this in :)

Issue #190, opened 2024-08-25
Modify Upgrade Message 


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In Upgrade currently the message looks `Scribe-Data vScribe-Data v3.3.0 (Update available: vScribe-Data v3.0.1)\nTo update: pip scribe-data --upgrade` .

Shouldn't it look like `Scribe-Data v3.3.0 (Update available: v3.3.1)\nTo update: pip scribe-data --upgrade`
in [here](https://github.com/scribe-org/Scribe-Data/blob/a072b52ba4ab627cf32714dd6587b0861ff1feb5/src/scribe_data/cli/version.py#L59)
If it is. @andrewtavis  Can i work on this?

andrewtavis:
Sounds good, @axif0! Feel free to change `Update` to `Upgrade` as well :)

axif0:
@andrewtavis 
Should I add cli test cases for version check? if yes, then where should I add it?

I'll change the version message to `Scribe-Data v3.3.0 (Update available: Scribe-Data v3.3.1)\nTo update: pip scribe-data --upgrade`

andrewtavis:
Closed by #191 🚀 Thank you for opening all of the issues today and getting to the fixes for them so quickly, @axif0! 😊

Issue #188, opened 2024-08-25
function name upgrade() to upgrade_cli() in upgrade.py


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In [main.py](https://github.com/scribe-org/Scribe-Data/blob/a072b52ba4ab627cf32714dd6587b0861ff1feb5/src/scribe_data/cli/main.py#L31) we import `upgrade_cli`, but in 
[upgrade.py](https://github.com/scribe-org/Scribe-Data/blob/a072b52ba4ab627cf32714dd6587b0861ff1feb5/src/scribe_data/cli/upgrade.py#L35) we called `upgrade()` is it .
For that i got error.
![image](https://github.com/user-attachments/assets/1d944581-66db-4159-a90b-a5e6f4a617bd)

andrewtavis:
Hey @axif0 👋 Do you want to send along a PR for this?

andrewtavis:
Thanks so much for the issue!

andrewtavis:
Looks like @pamarthiabhinav already sent along a change 🚀 I'll check this later today :)

axif0:
Ah ! already someone Already make an PR!

axif0:
@andrewtavis, in version currently the message looks`Scribe-Data vScribe-Data v3.3.0 (Update available: vScribe-Data v3.0.1)\nTo update: pip scribe-data --upgrade` .

Shouldn't it look like `Scribe-Data v3.3.0 (Update available: v3.3.1)\nTo update: pip scribe-data --upgrade`

Can i work on this?

andrewtavis:
Yes, @axif0, that makes sense :) You can also change `Update` to `Upgrade` :)

andrewtavis:
Thank you!

pamarthiabhinav:
@axif0 Apologies if you're working on this. But thanks for giving me a start for my contribution to the Scribe-Org.

pamarthiabhinav:
@andrewtavis PR #189 is ready for review. Check it out!

axif0:
[...]

PR #170, opened 2024-07-10
Add test for CLI tool

mhmohona:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

image

### Related issue

- #ISSUE_NUMBER

PR #191, opened 2024-08-25
fix: version message, add test for version

axif0:
### Contributor checklist

---

### Description

- Fix version message
- Add tests for Version Check. with command `pytest tests
` 

### Related issue

- #190 
- #188

### DIFF
--- a/tests/cli/test_version.py
+++ b/tests/cli/test_version.py
@@ -75,5 +75,5 @@ class TestVersionFunctions(unittest.TestCase):
         """
         Tests the scenario where a newer version is available, suggesting an update
         """
-        expected_message = "Scribe-Data v1.0.0 (Update available: Scribe-Data v1.0.1)\nTo update: pip scribe-data --upgrade"
+        expected_message = "Scribe-Data v1.0.0 (Upgrade available: Scribe-Data v1.0.1)\nTo update: pip scribe-data --upgrade"
         self.assertEqual(get_version_message(), expected_message)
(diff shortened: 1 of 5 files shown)


mhmohona (approved):
Hello @axif0, looks like you have addressed the review in https://github.com/scribe-org/Scribe-Data/pull/170#issuecomment-2282873796. Thanks for your great work! ✨ 

@andrewtavis, do you think for each command we should have separate test file like Asif has proposed?

andrewtavis:
Yes, I think that the way that @axif0 has it set up works well :) If the other tests in #170 could be split and added to `tests/cli` like the one here that would be great!

andrewtavis:
Thanks for the work here both of you! 😊

tests/cli/test_version.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @patch("scribe_data.cli.version.get_local_version", return_value="1.0.0")
+    @patch(
+        "scribe_data.cli.version.get_latest_version", return_value="Scribe-Data v1.0.1"
+    )
+    def test_get_version_message_update_available(
+        self, mock_latest_version, mock_local_version
+    ):
+        """
+        Tests the scenario where a newer version is available, suggesting an update.
```
andrewtavis, line 77: Very minor thing here that is more a personal thing for me, @axif0, I added the periods here as Python standards state that comments that are their own line should have a period at the end (and should also start with a capitalized work - they're full sentences). For inline comments it's best to not capitalize the first word and not end in a period :)

For next time! 😊

andrewtavis (approved):
Really appreciate the changes here, and especially adding in the tests, @axif0! Great work! Thanks for all your contributions today :) :)

PR #193, opened 2024-09-02
data-types path add

axif0:
### Contributor checklist

---

### Description
`data-types` Should be taken from [Here](https://github.com/scribe-org/Scribe-Data/blob/ac64c18483be32b8e553ea8b4b5a766c703d5517/src/scribe_data/resources/data_type_metadata.json#L2).


andrewtavis (approved):
Makes a ton of sense, @axif0! Really appreciate the care that you're putting into making sure we're reusing things :)

PR #195, opened 2024-09-02
minor redundant code fix

axif0:
### Contributor checklist

---

### Description
We can reuse [cli_utils](https://github.com/scribe-org/Scribe-Data/blob/9e9ceac3158deea011752c02f9fef4e01705b7c1/src/scribe_data/cli/cli_utils.py#L29) code in [list.py](https://github.com/scribe-org/Scribe-Data/blob/9e9ceac3158deea011752c02f9fef4e01705b7c1/src/scribe_data/cli/list.py#L29)

### DIFF
--- a/src/scribe_data/load/data_to_sqlite.py
+++ b/src/scribe_data/load/data_to_sqlite.py
@@ -328,7 +328,6 @@ def data_to_sqlite(
             )
 
     print("Database creation/update process completed.")
-    print("Database creation/update process completed.")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


src/scribe_data/load/data_to_sqlite.py
```diff
@@ -324,7 +324,6 @@ def table_insert(data_type, keys):
             )
 
     print("Database creation/update process completed.")
-    print("Database creation/update process completed.")
```
andrewtavis, line 327: Nice catch!

andrewtavis:
Can you check out the tests here, @axif0 :) I'm not sure that we can import variables from another files that are defined with `with ... as ...`. Maybe you could remove those and just have this be to remove the repeat print statement? Happy to hear your thoughts though!

axif0:
@andrewtavis  My bad I forgot to check the tests. Can you please check it?

andrewtavis (approved):
All seems to be working well now, @axif0! Appreciate all the cleanup you're doing! 🚚😊

Issue #197, opened 2024-09-03
Make paths in test_check_query.py Windows compatible


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

This issue is to use pathlib to make all paths in [test_check_query.py](https://github.com/scribe-org/Scribe-Data/blob/d36ed210d52582374a09f3dbe741cf17de0a939e/tests/wikidata/test_check_query.py) compatible with Windows. We should check the way that the paths are made with pathlib in other parts of the project, and then we can implement such that the pytest tests will pass :)

Happy to assist with this and review when it's ready!

CC @axif0 :)

axif0:
On it.. 🔥

andrewtavis:
Closed by #198 🥳 Thanks for your hard work, @axif0! Really has been nice to have you more involved :)

PR #198, opened 2024-09-04
Fix paths in test_check_query.py Windows compatible

axif0:
### Contributor checklist

---

### Description
Added a function `normalize_path` . Here it takes a file system path as an input and returns a "normalized" version of that path.
image

### Related issue

- #197

### DIFF
--- a/tests/wikidata/test_check_query.py
+++ b/tests/wikidata/test_check_query.py
@@ -25,7 +25,7 @@ from http import HTTPStatus
 from pathlib import Path
 from unittest.mock import MagicMock, mock_open, patch
 from urllib.error import HTTPError
-
+import os
 import pytest
 from scribe_data.wikidata.check_query.check import (
     all_queries,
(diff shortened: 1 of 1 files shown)


andrewtavis:
Hey @axif0 👋 Generally is looking good and I really like the inclusion of marks! Thanks for helping to organize the code :)

Would it be possible to leverage [pathlib](https://docs.python.org/3/library/pathlib.html) for this instead of os. In this way it'd be consistent with the rest of the project :)

axif0:
@andrewtavis thank you so much. pathlib is more realistic to use here. 

@mhmohona In my windows environment, I got no error in testing, for double checking can you please verify it by running the wikidata test ?

andrewtavis:
Thanks so much for the quick fixes here, @axif0! :)

andrewtavis (approved):
Thanks for the quick reaction to the comments, @axif0! @mhmohona, do let us know if you're seeing anything that needs to be done about the recent commits. For now we're good to merge 😊

mhmohona:
> Thanks for the quick reaction to the comments, @axif0! @mhmohona, do let us know if you're seeing anything that needs to be done about the recent commits. For now we're good to merge 😊

@andrewtavis I just checked. It seems things are working fine, without any error on Windows. ✨ 

and @axif0, good work. 👏🏾

Issue #200, opened 2024-09-19
--Interactive cli command error


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Testing Interactive Cli command, I find the error. Is the command correct? 

![image](https://github.com/user-attachments/assets/ab95d156-ce53-47f8-ba7c-3ca0b1dd8ab2)

andrewtavis:
There is at least one bug with interactive mode right now, @axif0, but as of yet no issue. Is this similar to what you were seeing, @mhmohona?

mhmohona:
Thank you @axif0 for bringing it up. Yes @andrewtavis, its one of the errors. Submitting PR right away.

andrewtavis:
Closed by #201 🚀 200 issues and PRs! Thanks so much for the fix, @mhmohona, and similarly for the issue to document the changes, @axif0 😊

PR #201, opened 2024-09-20
fix interactive bug

mhmohona:
### Contributor checklist

---

### Description
Fixed loop, path and output type error.

image

### Related issue

Fixes - #200


andrewtavis (approved):
Thanks for the quick fix, @mhmohona! Let's look into testing for interactive mode and a few other parts of the CLI that might still need it :) Feel free to make issues that you think would address some of the vulnerabilities 😊

Issue #202, opened 2024-09-21
Interactive cli issue


axif0:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Seams like `Overwrite existing data without asking (y/n) [n]:` not working. May I know what it is for? 

![image](https://github.com/user-attachments/assets/db45b22d-4a14-460d-b512-be026f6aa1f7)

also `(a) for all languages` is occurring a bit of confusion for me :(

![image](https://github.com/user-attachments/assets/02f70934-712a-482e-ab12-ddcc7b990339)

axif0:
I want to work on it, if it is not problem .

andrewtavis:
CC @mhmohona for the discussion here :) The overwrite for existing data is to make sure that we're not overwriting already existing Scribe data if maybe the user needs it. And yes you can investigate the issues with `a`, @axif0!

andrewtavis:
Post #204 we're quite close on the interactive functionality and the CLI in general. Can we take a look at `a` for all languages and data types here before we jump to #203? I think it'd be good for us to be able to say that it's stable and tested before jumping to making it fancier :)

andrewtavis:
Let's make sure we can go through all of [USAGE.md](https://github.com/scribe-org/Scribe-Data/blob/main/USAGE.md) error free and then we can close this! I'll check it out a bit more over the weekend as well.

andrewtavis:
Closed by #268 🚀 Great PR, @axif0!

PR #204, opened 2024-09-24
Fix more bug in CLI - Interactive Learning module 

mhmohona:
### Contributor checklist

---

### Related issue

- #202

### DIFF
--- a/src/scribe_data/cli/convert.py
+++ b/src/scribe_data/cli/convert.py
@@ -29,25 +29,23 @@ from typing import Optional
 from scribe_data.cli.cli_utils import language_map
 from scribe_data.load.data_to_sqlite import data_to_sqlite
 from scribe_data.utils import (
-    DEFAULT_JSON_EXPORT_DIR,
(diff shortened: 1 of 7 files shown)


mhmohona:
This change in `test_get.py` file has been made to include the `overwrite` parameter in the test's expected calls so they match the actual calls to `query_data`, resolving the `AssertionError` caused by mismatched arguments in CI test.

andrewtavis:
Minor changes in 4cbc502 as I felt like the CLI was a bit too verbose. It was saying it was going to do something multiple times. The path for exporting JSONs also wasn't working properly as it needs to be absolute. @mhmohona, do you want to take a look at some of these changes and we can discuss how to go from here? Multiple data types is also working, so we're generally good in those regards :)

andrewtavis (approved):
Definitely some progress on interactive mode for this one, @mhmohona! Thanks for the continued efforts 😊 I'm thinking we're really close on it and the cli in general :)

PR #205, opened 2024-10-01
Completed Hebrew Queries

DanielGevaZA:
### Contributor checklist

---

### Description

- #ISSUE_NUMBER


andrewtavis (approved):
Quick note @DanielGevaZA that the Hebrew verbs query wasn't working in the way that it was sent along. Would have been good to test it on the [Wikidata Query Service UI](https://query.wikidata.org/) to see that it wasn't returning most conjugations. I've updated it now :) With this comes a bunch of `MARK` notices throughout the queries as I felt it helped with readability of the new queries, and decided to do it for all of the old ones 😊

Thank you!

Issue #219, opened 2024-10-03
Encountering IncompleteRead exception when making HTTP requests


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

IncompleteRead error occurred when I was trying to query data from Wikidata using the SPARQLWrapper library. If I try all languages along with data-type it causes me the error in interactive mode.

![image](https://github.com/user-attachments/assets/1d824117-697b-4679-a40f-b6e797b4689a)

I'll be working on this. pr on the way.

andrewtavis:
Closed by #221 😊 Thanks for the issue and the work here, @axif0!

PR #221, opened 2024-10-03
fixing IncompleteRead issue

axif0:
### Contributor checklist

---

### Description
Adding this retry logic, we can handle temporary network issues or server-side problems that cause the IncompleteRead error. The code will automatically retry the request a few times before giving up.

### Related issue

- #219


andrewtavis:
Hey @axif0 👋 Can you look into the merge conflicts here? :) One thing as well: I'm a bit confused why this is necessary as we already had something to retry queries, but maybe the old one wasn't working 🤔 You can see the old version [here](https://github.com/scribe-org/Scribe-Data/blob/02c7e1bca7c09649ca63b2b3ea3ba78fda3625f8/src/scribe_data/wikidata/query_data.py#L171). Do you want to fix the conflicts and combine the methods as you see fit?

axif0:
@andrewtavis  While running all language querying verbs and nouns, I got the `IncompleteRead` error specifically. After that I ran again and no error. I thought if the error occurs then try few times to reconnect. Maybe some network connection issue from my end. Sorry if you find it irrelevant.

andrewtavis:
No stress whatsoever, @axif0! It's an interesting idea. I'm just wondering if maybe your incomplete read is a more specific implementation of what I was doing with mine. In that case, maybe you could do a slight edit to combine them? :)

andrewtavis:
Let's not forget this, @axif0 :) Basically your implementation checks for `IncompleteRead`, which is what we get back when maybe we've been rate limited by the Query Service or something similar, and then it runs the queries again a certain number of times?

axif0:
@andrewtavis ya.. Agreed with that. I think updated logic is enough. 😅 New implimentation should handle `IncompleteRead` effectively. we can close this PR.

andrewtavis:
I think this is maybe a good way of combining the approaches, @axif0 :) We'll get a more explicit error now, and we'll also let the user know that the query will be retried as your approach did 😊

andrewtavis (approved):
Thanks as always for the care you're putting into the project, @axif0!

Issue #220, opened 2024-10-03
Create Scribe-Data Pidgin data process queries


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would create the queries for [Pidgin languages](https://en.wikipedia.org/wiki/Pidgin) in the [src/scribe_data/language_data_extraction/Pidgin](https://github.com/scribe-org/Scribe-Data/tree/main/src/scribe_data/language_data_extraction/Pidgin) directory. To start we can make a nouns query and a verbs query in two separate PRs, and from there we can make new issues for other types of data. These queries can be based on the already existing queries for other languages 😊

### Contribution

Happy to support with this! Please let me know if there are questions, and I can also review once a PR is ready :)

DeleMike:
Hi @andrewtavis I would love to work on this, please.

DeleMike:
But what kind of pidgin? There are [different types of pidgin](https://en.wikipedia.org/wiki/Pidgin#List_of_notable_pidgins)
I am familiar with [Nigeria Pidgin](https://en.wikipedia.org/wiki/Nigerian_Pidgin)

andrewtavis:
Hmmmmm, great point, @DeleMike :) You can work on Nigeria Pidgin, but then should we name the directory this as well?

DeleMike:
Yes, great point too @andrewtavis!  To avoid ambiguity, we should do that.

We could have:
```bash
language_data_extraction/
    Nigerian_Pidgin/
    Other_Types_Of_Pidgin/
```

what do you think? If you agree, do we change the issue name or create a new one?

andrewtavis:
Let's first also check to see which versions on Pidgin have data on Wikidata with [src/scribe_data/check_language_data.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/check_language_data.sparql). I'd say that the best naming criteria would also maybe be `Pidgin_Nigerian` so that in the directory structure all of the Pidgin versions will be together. We can also just create individual directories for individual types of Pdigin as is needed :)

DeleMike:
Yes, you are right! We should use that convention and thanks for the link!

[...]

PR #246, opened 2024-10-04
feat(nigerian-pidgin): add SPARQL query for verbs

DeleMike:
- Create new directory structure for Nigerian Pidgin
- Add basic SPARQL query to extract Nigerian Pidgin verb lemmas from Wikidata

This PR introduces initial support for Nigerian Pidgin verb data extraction, providing a foundation for gathering basic linguistic information for this language.

### Contributor checklist

---

### Related issue

- #220


DeleMike:
Hi @andrewtavis👋🏾, here is the PR that adds verbs query for Nigerian Pidgin. 

It is related to #220.

Questions:
1. Do we add nouns query in this PR or keep it lean like this for now?

2. I saw that there is a directory called:
```bash
scribe_data_json_export/English
``` 
I believe that is for English but do we need to create something similar for `Pidgin_Nigerian`? How would that work?

andrewtavis (approved):
I was thinking about this a bit based on our conversations, @DeleMike, and I think having a sub-directory structure for these sorts of languages makes sense so that we can have dialects structured in a similar way. What I mean by this is having the different types of Norwegian in a single directory, or for something like Portuguese we could have Portuguese and Brazilian variants :) I'll merge as it is, but also let me know what your thoughts are on this! We can maybe add more versions of Pidgin into this new directory?

DeleMike:
hi @andrewtavis, this makes sense! I was going through the updated code and this will make sure we have things in uniform. 

I think your update is what we should go for!

Issue #225, opened 2024-10-03
Expand Danish verbs query


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would look into expanding the [src/scribe_data/language_data_extraction/Danish/verbs/query_verbs.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/language_data_extraction/Danish/verbs/query_verbs.sparql) file with as many conjugations as are possible from the current verbs on Wikidata. We can use code for getting verbs from other languages, and from there we can check Danish verbs on Wikidata for what conjugations are available. We can then expand the query with optional selections of certain conjugations as is done in other SPARQL queries. The query can be tried on the [Wikidata Query Service UI](https://query.wikidata.org/) during development :)

### Contribution

Happy to support as I can with answers to questions as well as code review once the PR is up!

catreedle:
Hi @andrewtavis, I would like to work on this issue :)

andrewtavis:
Sounds great, @catreedle! I'll check the PR later :)

andrewtavis:
Closed by #249 😊 Thanks so much, @catreedle! 🚀

PR #249, opened 2024-10-05
Expand Danish verb query

catreedle:
### Contributor checklist

---

### Description

This pull request expands the Danish verb query in [src/scribe_data/language_data_extraction/Danish/verbs/query_verbs.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/language_data_extraction/Danish/verbs/query_verbs.sparql) by adding more conjugation forms available on Wikidata.

The forms included in this update are:

- Infinitive: The base form of the verb.
- Infinitive Active: The active form of the infinitive.
- Present Tense Active: Verb form indicating an action currently happening.
- Preterite Active: Simple past tense form for completed actions.
- Past Participle: Used in perfect tenses or as an adjective.
- Present Participle: Non-finite form used for continuous tenses or as an adjective.
- Imperative: Verb form used for commands.
- Present Passive: Present tense form indicating passive action.
- Preterite Passive: Passive form of the past tense.
- Infinitive Passive: Passive form of the infinitive.

The query has been tested and runs correctly on the [Wikidata Query Service](https://query.wikidata.org/)

### Related issue

- #225

### DIFF
--- a/src/scribe_data/language_data_extraction/Danish/verbs/query_verbs.sparql
+++ b/src/scribe_data/language_data_extraction/Danish/verbs/query_verbs.sparql
@@ -2,7 +2,7 @@
 # All Danish (Q9035) verbs and the currently implemented tenses for each.
 # Enter this query at https://query.wikidata.org/.
 
-SELECT
+SELECT DISTINCT
   (REPLACE(STR(?lexeme), "http://www.wikidata.org/entity/", "") AS ?lexemeID)
   ?infinitive
   ?infActive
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
This is really nice, @catreedle :) I removed select distinct and I don't think it'll do much to avoid the repeat rows being returned and slows the query down a bit. We can revisit it if need be later 😊

andrewtavis:
Select distinct will add an extra processing step to make sure all things are unique. Generally for this we can expect things to be unique, or if they aren't then there's a problem with the modeling that means we're getting multiple values back for certain form combinations. Big thing is to not use extra resources for the distinct of not necessary :)

PR #252, opened 2024-10-05
Path fix

axif0:
### Contributor checklist

---

### Description
In windows the path works even if it contains spaces. Like it can handle "New folder" now. Previously it shows error like- 

`python3: can't open file 'E:\New': [Errno 2] No such file or directory
`

### DIFF
--- a/src/scribe_data/wikidata/query_data.py
+++ b/src/scribe_data/wikidata/query_data.py
@@ -220,7 +220,7 @@ def query_data(languages=None, word_types=None, overwrite=None):
 
             # Call the corresponding formatting file.
             os.system(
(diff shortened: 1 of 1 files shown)


axif0:
If someone in windows faced like 
`ModuleNotFoundError: No module named 'scribe_data'
`
Please use like `set PYTHONPATH=E:\New folder\Scribe-Data\src`.

andrewtavis:
I felt like moving the path to a separate variable was a bit cleaner, @axif0 :)

andrewtavis (approved):
Thanks as always for finding bugs and fixing them so quickly, @axif0!

PR #256, opened 2024-10-05
Expand Arabic Nouns SPARQL Query to Include Additional Forms

OmarAI2003:
### Contributor checklist

---

This PR enhances the existing SPARQL query by including new grammatical forms for Arabic nouns. Previously, the query fetched only singular, plural, and dual forms. Now it has been expanded to also include:

- Definite singular
- Definite plural
- Feminine singular
- Feminine plural

**Linguistic Context:**
The added forms are crucial for capturing Arabic morphology more accurately. The definite and feminine forms were chosen because they are the primary variations used in formal and standard Arabic. These forms affect the meaning and usage of nouns significantly, especially in text corpora where definiteness and gender distinctions are key.
make it a small bit short

As a result, the number of fetched lexemes increased from **589** to **1945**. This will provide a more comprehensive dataset for Arabic noun lexemes, making it suitable for more detailed linguistic analysis.


OmarAI2003:
I noticed that there is a duplicate commit in this PR [here ](https://github.com/scribe-org/Scribe-Data/pull/256/commits/b7dd4833e986d3dc471add55db45e201dff72811)from my previous PR. This occurred because I based this branch on the earlier one. Please consider reverting that duplicate commit in this PR. Please let me know your preference, and I'll make the adjustments. Thank you!

andrewtavis:
I think we're generally fine here, @OmarAI2003. We can just merge this and then merge the other one after so that the edits to the adjective file aren't in conflict :) I'll check and merge this one in when the edits to the other PR go through!

So happy to have edits to the Arabic queries going through! Hopefully the data is expanding 😊

andrewtavis:
One thing here, @OmarAI2003: It seems like for the nouns we're getting lots of repetition. Do we want to split them based on genitive, accusative and other forms as well so we don't get multiple results for each?

OmarAI2003:
See here the same problem with nouns some popular nouns like (mother) (أم) have many forms, while other non-popular nouns like (spirit) (روح) have limited forms.
a
b

andrewtavis:
Maybe it makes sense to model for the 21 forms such that we make the assumption that the other nouns will get there eventually? It's not like these forms just don't exist, and that way we won't need to rewrite the query later?

OmarAI2003:
Hi gain @andrewtavis
Thanks for your feedback on the PR! I’ve put a lot of effort into revising the query to ensure that each lexeme form is unique by incorporating all relevant grammatical features.My approach aims to eliminate duplicates while providing a comprehensive representation of Arabic nouns.

OmarAI2003:
I think these PRs #256, #255 are ready for merging now, @andrewtavis. However, the verb queries have so many duplicates, and the query is split across multiple files and is so complicated. Does that seem like a serious issue I can work on?

andrewtavis (approved):
Similar to the adjectives, @OmarAI2003, let's get what we can for now and we can expand it later :) There's not much data now, but this generally gets unique entries for a lot of what's there.

Your feedback's very welcome here! One thing to note, we really do need the base output types to be consistent, so the result in any column or JSON output value needs to be the label of a form, not the props of the form or anything else.

Hope this makes sense!

Issue #209, opened 2024-10-03
Add functionality to re-ask user for an option to the Scribe-Data interactive mode


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The Scribe-Data interactive mode can be accessed in the following way:

```bash
scribe-data get -i
```

With this the user is prompted to select options that would be appropriate to get the data that they need. One thing that's missing is functionality to re-ask the user for a response if they did not provide one. If the user provides an invalid response, then the program should error, but if the user just hits enter then they should be asked the same question again - i.e.:

```bash
$ scribe-data get -i
Welcome to Scribe-Data interactive mode!

Language options:
1. English
2. French
3. German
...

Please enter the languages to get data for, their numbers or (a) for all languages:

# User hits enter, and what should follow is...

Language options:
1. English
2. French
3. German
...

Please enter the languages to get data for, their numbers or (a) for all languages:
```

### Contribution

Happy to support this with a review and answer any questions a contributor might have!

KesharwaniArpita:
@andrewtavis I have an idea for this. I think if we update the interactive.py with the following changes we will be able to add the functionality to re-ask the user for a response if they haven't provided one:
1. Modify the `get_selection()` function to re-ask for input if no valid input is provided.
2. Add a loop in the `select_languages()` and `select_data_types()` functions to handle re-asking the question.
3. Make the output options prompt also re-ask the user in case of an empty response.

andrewtavis:
Sounds great, @KesharwaniArpita! Let's finish the other issue and then we can move on to this one :)

omoladd:
Hi @andrewtavis  , I'm interested in this. Can you assign this to me?

KesharwaniArpita:
Hi @omoladd, I was very curious about how to solve this issue and tried to tweak the existing functions a little bit in the day. Since you also seem interested would you like to take a look into the PR and see if it can be improved your suggestion would mean a lot.

omoladd:
@KesharwaniArpita, I didn’t realize you already responded. 
I opened a PR too; https://github.com/scribe-org/Scribe-Data/pull/239

KesharwaniArpita:
[...]

Issue #203, opened 2024-09-23
Investigate and POC a terminal UI for `--interactive` CLI mode


wkyoshida:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Creating the issue to formalize the idea brought in https://github.com/scribe-org/Scribe-Data/issues/158#issuecomment-2218847941

---

This would be for an evolution of the `--interactive` mode of the CLI to implement now with a TUI - meaning along the lines of:

![image](https://github.com/user-attachments/assets/bd99d3f4-016e-4d59-91c7-b11d3aa04389)

where the CLI expands and takes up the terminal with an interface.

Some preliminary references found for this:
- [Python module for the curses library](https://docs.python.org/3/library/curses.html)
- [Guide for the Python curses module](https://docs.python.org/3/howto/curses.html)
- [StackOverflow - console application gui for python](https://stackoverflow.com/questions/12664184/console-application-gui-for-python)
- [Cross-platform Python package for text UIs](https://github.com/peterbrittain/asciimatics) *this one appears particularly interesting

Some points to answer/discuss:
- Any major limitations with the approach?
- Would we want this to replace the current `--interactive` mode or instead be its own additional mode? A `--pretty-interactive` if you will or something of the sort..

### Contribution

Happy to support someone interested in this 😁

--- 

CCing those who contributed to the `--interactive` mode so far, in case there's any interest:
@mhmohona and @axif0

mhmohona:
It would be really cool! This way user will get more visual, making things easier to use. 

Right now I am working on the recent issue @axif0 has raised. After that I can work on it. Or @axif0, if you want, you can pick it up as well.

axif0:
Yaa, I can try

andrewtavis:
Might make sense for you to map out your ideas for this one after a quick investigation, @axif0 :) Let us know how we can support!

andrewtavis:
Closed by #268 🚀 Really looking great, @axif0! Let's test it a bit more and make any issues that are needed :)

PR #239, opened 2024-10-03
feat: add functionality to re-ask user for an option to the scribe-da…

omoladd:
…ta interactive mode

### Contributor checklist

- [X] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

 added functionality to re-ask user for an option to the scribe-data interactive mode

### Related issue

- #ISSUE_NUMBER
#209

PR #268, opened 2024-10-07
Interactive functionality

axif0:
### Contributor checklist

---

### Description
The [rich](https://github.com/Textualize/rich) library enhances CLI output with colorful tables and messages, while  [questionary](https://github.com/tmbo/questionary)  provides interactive prompts for user input.

attempt 2

### Related issue

- #203 
- #202
- #209


axif0:
cc: @mhmohona, @andrewtavis if you don't like the interactive UI or if it is not user friendly for community, I can make another PR using  [asciimatics](https://github.com/peterbrittain/asciimatics) as @wkyoshida said.

src/scribe_data/cli/interactive.py
```diff
@@ -23,168 +23,156 @@
...
+                    config.overwrite,
+                    config.output_type,
+                )
+                rprint(f"[green]✓[/green] Exported {language} {data_type} data.")
+
+    rprint(Panel.fit("[bold green]Data export completed successfully![/bold green]"))
+
+
+# MARK: Start interactive mode functionality
+def start_interactive_mode():
```
mhmohona, line 154: Please add docstring here, as well as for other functions.

mhmohona (changes requested):
Great work @axif0! ✨

andrewtavis:
One question on this, @axif0: What's the support from the packages you're using to display options as a grid? I'm quite conscious right now that as soon as we start adding in the new languages that people are working on, that we won't be able to see all the options on the screen anymore without displaying a grid. Do you think this is an option for the current dependencies?

Also, don't we need dependencies added to the `requirements.txt` file?

andrewtavis:
Sorry the dependencies are there... Lots of PRs over the last few days. I'm going to start testing this and comparing it against the work to re-ask users for options that was done in #216 and #239. I'm assuming that the dependencies will do this, but we'll see :)

andrewtavis (approved):
Ok lots of changes went through in my commit, but overall this is great, @axif0 😊

- Exit wasn't working
- We can now rerun data type only
- Minor color/style fixes
- Copy fixes (I changed export to request so it doesn't sound like the data is in Scribe-Data)
- Overall control fixes
- Asking the user if they want to continue when they didn't select something as the message was above the options for me, and is definitely going to be there for everyone as the options expand

This really is so cool :) :)

Issue #229, opened 2024-10-03
Expand Kurmanji data queries


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would look into expanding the [src/scribe_data/language_data_extraction/Kurmanji](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/language_data_extraction/Kurmanji) files with as much data as are possible from the current data on Wikidata. We can use code for getting data from other languages, and from there we can check Kurmanji data on Wikidata for what conjugations are available. We can then expand the query with optional selections of certain forms as is done in other SPARQL queries. The query can be tried on the [Wikidata Query Service UI](https://query.wikidata.org/) during development :)

Data types to include:
- [x] Nouns
- [x] Verbs
- [x] Adjectives
- [x] Adverbs
- [x] Prepositions
- [x] Emoji keywords

### Contribution

Happy to support the development and review when the PR is up 😊

Khushalsarode:
@andrewtavis I will work on this!

andrewtavis:
Thanks for picking it up, @Khushalsarode! Let us know if you need any support :)

Khushalsarode:
![wikidata](https://github.com/user-attachments/assets/c8d4a991-dab4-441d-8965-047af591845f)
@andrewtavis 
singulars

Khushalsarode:
@andrewtavis would you like me to work on Marathi entity verbs/prepositions on wikidata?

andrewtavis:
That's kind of up to you, @Khushalsarode :) If you'd like to expand them, then that would be really cool! I'm not sure if we can count Wikidata edits for Outreachy or Hacktoberfest, but if you're interested in editing a bit, then by all means!

andrewtavis:
Just added a list of data types that we want to include to this issue :) Have marked those that are already done or have PRs open, and we can work on the others 😊 If the data type can't work, then we can move to the others and open up specific issues later :)

Khushalsarode:
Thanks! 
Can I work on German and Italian data query language, or should I go one by one.
Kind of feel I can do both with ease!
@andrewtavis

andrewtavis:
[...]

PR #272, opened 2024-10-07
adding query_verbs.sparql for kurmanji

Khushalsarode:
### Contributor checklist

### Description

- The query has been executed & tested on wikidata editor 
- Developed a SPARQL query to retrieve all Kurmanji (Q36163) verbs and their grammatical forms from Wikidata.
- Extracts lexeme IDs, verb forms, and various grammatical features including infinitive, gerund, accusative, and ergative.
- optional clauses to include additional forms such as adverbial, absolute construction, and participle

Issue reference
- #229 

@andrewtavis

### DIFF
--- a/src/scribe_data/language_data_extraction/Kurmanji/verbs/query_verbs.sparql
+++ b/src/scribe_data/language_data_extraction/Kurmanji/verbs/query_verbs.sparql
@@ -1,70 +1,14 @@
 # tool: scribe-data
-# All Kurmanji (Q36163) verbs.
+# All Kurmanji (Q36163) verbs and the currently implemented tenses for each.
 # Enter this query at https://query.wikidata.org/.
 
 SELECT
   (REPLACE(STR(?lexeme), "http://www.wikidata.org/entity/", "") AS ?lexemeID)
   ?verb
-  ?form
-  ?grammaticalFeature
-  ?infinitive
-  ?directCase
-  ?gerund
-  ?intransitivePhase
-  ?basicPhase
-  ?conjParticiple
-  ?adverbial
-  ?absConstruction
-  ?accusative
-  ?ergative
-  ?additivePhase
 
 WHERE {
   ?lexeme dct:language wd:Q36163 ;
     wikibase:lexicalCategory wd:Q24905 ;
     wikibase:lemma ?verb .
-
-  
-  # Infinitive (basic form of the verb)
-  FILTER(lang(?verb) = "ku") .
-
-    # Forms associated with the grammatical features (direct case, gerund, etc.)
-  OPTIONAL {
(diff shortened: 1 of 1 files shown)


src/scribe_data/language_data_extraction/Kurmanji/verbs/query_verbs.sparql
```diff
@@ -5,9 +5,78 @@
...
+    FILTER(LANG(?ergative) = "ku") .
+  } .
+  
+  
+  
+  
+
 }
+
+ORDER BY ASC (?lexemeID)
```
andrewtavis, line 79: Can we remove the extra spacing, @Khushalsarode? And also this ORDER BY is not necessary here as we don't need the results in a given order. This will end up just slowing the query down a bit :)

Khushalsarode: sure!

Khushalsarode: done as per requested! 👍

Khushalsarode: shall I mark "resolve conversation"?

Khushalsarode: Can I record this PR to outreachy!

andrewtavis (approved):
Thanks for this, @Khushalsarode :) Ultimately I simplified the data down as there there don't appear to be any forms on the verbs, so this is one we'll need to do later. Nice touch on filtering the language to `ku` though 😊

Khushalsarode:
shall I create separate query file for each in list?
@andrewtavis

andrewtavis:
In the list in the issue, @Khushalsarode? Yes, please :)

Khushalsarode:
> In the list in the issue, @Khushalsarode? Yes, please :)

also I tried query for emoji's but no data has been prompted, Queried the kurmanji language against the emoji entity.
so should such queries be pushed or not, in case if data may be available in future?

andrewtavis:
Ah, that's not how emojis work. We're not getting them from Wikidata, but rather from unicode. Check the emoji directories for other languages to see how it's done :) I'm not sure if Unicode has Kurmanji-emoji relationships though 🤔

Khushalsarode:
> Ah, that's not how emojis work. We're not getting them from Wikidata, but rather from unicode. Check the emoji directories for other languages to see how it's done :) I'm not sure if Unicode has Kurmanji-emoji relationships though 🤔

oh, I see my mistake. I checked the emoji directories for other languages. also read on internet that this unicode need to be only localize for different languages that doesn't change the unicode. nice!

PR #274, opened 2024-10-07
Dynamic PYTHONPATH and Cross-Platform Script Execution

axif0:
### Contributor checklist

---

### Description
If user give queries, windows likely to run [format_target_type.py](https://github.com/scribe-org/Scribe-Data/blob/f9e6f603db190e012dd956fc1b99944625482d54/src/scribe_data/language_data_extraction/English/nouns/format_nouns.py) without an environment. Here `execute_formatting_script` dynamically sets the `PYTHONPATH` and runs a formatting script using `subprocess.run` on either Windows or Unix-like systems

- [Calling OS' Python Executable](https://stackoverflow.com/questions/52356102/calling-os-python-executable-from-another-python-process)

### DIFF
--- a/src/scribe_data/wikidata/query_data.py
+++ b/src/scribe_data/wikidata/query_data.py
@@ -22,10 +22,10 @@ Updates data for Scribe by running all or desired WDQS queries and formatting sc
 
 import json
 import os
+import subprocess
+import sys
 from pathlib import Path
 from urllib.error import HTTPError
-import sys
-import subprocess
 
 from tqdm.auto import tqdm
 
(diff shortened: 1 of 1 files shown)


src/scribe_data/wikidata/query_data.py
```diff
@@ -262,7 +264,39 @@ def query_data(
...
+
+def execute_formatting_script(formatting_file_path, output_dir):
+    # Determine the root directory of the project
+    project_root = Path(__file__).parent.parent
+
+    if sys.platform.startswith("win"):
+        python_executable = sys.executable
+        pythonpath = str(project_root)
+
+        # Create environment with updated PYTHONPATH
```
andrewtavis, line 281: No stress now, @axif0, but can we get periods at the end of full line comments in the future. Pythonic-wise they should be complete sentences :) I'll fix this locally and send it along shortly!

andrewtavis (approved):
Adding in docstrings for functions is also really important, @axif0, as that means that the docs are consistently being updated by autodoc 😊

Thanks for your support for the community and getting to these key fixes so quickly!

PR #279, opened 2024-10-08
Created folder for Punjabi language and added query

SethiShreya:
Signed-off by: Shreya Sethi
shreyasethi47@gmail.com

- Create folder for the Punjabi language
- Added Sparql query for Punjabi nouns
- Added filter to only get Gurmukhi language not Shahmukhi
- Query include gender too


andrewtavis (approved):
Thanks for the work here, @SethiShreya! Let me know what your thoughts are on my question in the issue and we can go from there for other queries 😊

Issue #142, opened 2024-06-07
Simplify formatting process to lexeme based outputs rather than string based


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would work on simplifying the data formatting process - specifically the noun formatting processes - such that they exports lexeme based files from the data. As of now the formatting processes are very long and migrate all the data from Wikidata into string based key JSONs, so lexemes with different meanings but that are the same string would be in the same entry. We do not want to do this anymore based on decisions in #59 and #110 :)

- Note: the Swedish nouns process should also be improved such that definite versions of nouns are also included :)

### Contribution

I'll be working on this, and once it's done the old state of Scribe-Data will be officially cleaned up!

andrewtavis:
Closed by #514 😊

PR #286, opened 2024-10-08
Fix HTTP Error 405 in Hindi Language

axif0:
### Contributor checklist

---

### Description
Commenting only 3 lines out of 4 on top of the file, fix's the `HTTP Error 405: Method Not Allowed`. 

CC: @andrewtavis @SethiShreya


andrewtavis:
Hey @axif0 👋 No stress on this, but I'm going to close this as the formatting is going to be having an overhaul soon. I need to do #142 soon that will change how formatting works entirely. Once this is done we can look into this again, but I'd prefer to not have to navigate more legacy formatting files than are already there.

Hope this makes sense, and again no stress!

PR #310, opened 2024-10-10
Ar adverbs

OmarAI2003:
### Contributor checklist

---
### Description
This PR adds a file to extract adverbs using a SPARQL query. Currently, the Arabic directory supports extracting nouns, verbs, adjectives, and now adverbs. What should be the next step?


andrewtavis (approved):
Thanks for this, @OmarAI2003! Big thing is that we don't need the label service for this as it'll come from the lemma. It makes more sense to not include it as including it does add to the computation load :)

Issue #295, opened 2024-10-09
Proposal for improving CLI Total functionalities 


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

For total lexemes we need to put **language QID** along with **data-type QID**. So both QID is statically putted on language_metadata and data_type_metadata.

Like when user put English and nouns (if wants) then from metadata QID passes in the query.
For example, If user passes Hindi and uses nouns, then as the [language_metadata](https://github.com/scribe-org/Scribe-Data/blob/e2ce7d136b885c0b12b8bd2cc92ee327ccf31874/src/scribe_data/resources/language_metadata.json) does not has the QID, then it returns empty and as [data_type_metadata](https://github.com/scribe-org/Scribe-Data/blob/e2ce7d136b885c0b12b8bd2cc92ee327ccf31874/src/scribe_data/resources/data_type_metadata.json) has data-type therefore the query returns error total lexems.

- [ ] Add a check function that check if a language or data-type are in the [language_metadata](https://github.com/scribe-org/Scribe-Data/blob/e2ce7d136b885c0b12b8bd2cc92ee327ccf31874/src/scribe_data/resources/language_metadata.json) or  [data_type_metadata](https://github.com/scribe-org/Scribe-Data/blob/e2ce7d136b885c0b12b8bd2cc92ee327ccf31874/src/scribe_data/resources/data_type_metadata.json).
-  [ ] User can simply put QID for both language and data-type. like `scribe-data total -l Q1860 -dt Q1084` (ex. english, nouns).
- [ ] We need to add functionalitys like `scribe-data total -lang LANGUAGE -a` to loop through to allow for --total to take a list and show all the available total lexemes.
- [ ] Total functionality should can handle specific language are   `scribe-data total --check-data -l Q1860` (for english) or  `scribe-data total --check-data -dt Q1084` (for nouns). 
- [ ] After all the changes made, we can add the total functionalities on interactive cli mode.

CC: @andrewtavis @mhmohona Not sure if the ideas are truly needed. 

### Contribution
Happy to work on this and support someone who has interest!

mhmohona:
[...]

PR #311, opened 2024-10-10
Expanded total command functionality 

SethiShreya:
Signed off by Shreya Sethi
shreyasethi47@gmail.com

Expanded total command to include
```scribe-data total --all``` 
=> to list down lexemes of all languages and data types in Scribe

```scribe-data total --language Language``` 
=> to list down all the data types of input language

```scribe-data total --language Qid``` 
=> to list down all the data types of language with qid

```scribe-data total --language Lang(or qid) -dt datatype``` 
=> print lexemes of given language and datatype

Tested all these commands on my side

Issue #293, opened 2024-10-09
Update languages metadata file and use of it thoughout project


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As of now the Scribe-Data CLI options are determined based on the [language_metadata.json](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/language_metadata.json) file. To make maintenance of the package easier, it would be great if the options of the CLI were instead determined by the directory structure of [src/scribe_data/language_data_extraction](https://github.com/scribe-org/Scribe-Data/tree/main/src/scribe_data/language_data_extraction) so that the code doesn't need to be updated each time new queries are being added in.

Of key importance is also that the options of the CLI would allow for dialects as well, so for Norwegian we'd like to see `Norwegian - Bokmål` and `Norwegian - Nynorsk`, for example. How this will be achieved is open for discussion!

### Contribution

Happy to discuss how best to read in dialect sub directories and review the changes here when the PR is up!

OmarAI2003:
I'm interested in this issue. 😃

andrewtavis:
This would be a really good one for you, @OmarAI2003 😊 Let us know if you have any questions!

andrewtavis:
In talking about this a bit, @OmarAI2003, we might not be able to do this. @SethiShreya and I were talking and as you said we need the QIDs as well so that we can do calls for the CLI based on QIDs as well. Without a central store of languages and their QIDs, maybe it can't work?

OmarAI2003:
Maybe we could use the directory structure just for language names, but still keep `language_metadata.json` for properties like QIDs? Not sure if this would help, but happy to hear your thoughts!

andrewtavis:
Is an interesting idea, but then say that we rely on the structure and then we don't get a QID added and then some functionality is broken 🤔

OmarAI2003:
So is this issue will be closed , or is there anything that needs to be addressed?

andrewtavis:
[...]

PR #330, opened 2024-10-12
fix: refiine cli to resolve languages data fetch incosistency

### LINKED ISSUE #328: Refine CLI User Experience by Validating Input Languages and Data Types
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Summary
When users provide a non-existent language or data type to the total command, the system incorrectly returns a number of lexemes. This leads to confusion and undermines the user experience.

### Steps to Reproduce

1. Run the total command with an invalid language or data type.
2. Observe that the command returns a number rather than indicating that the language or data type does not exist.

### Example
Initially when we run this command `scribe-data t -lang Latin` and we had:
```bash
lang filter =  ?lexeme dct:language ?language . # added this while trying to debug
Language: Latin # and this...
Total number of lexemes: 1344820
```
After we added some print statements, we see that the language_filter was not updating the `language` parameter hence giving a wrong result.

You can see the same thing for French: ` scribe-data t -lang French`, and we had:
```bash
Lang filter =  ?lexeme dct:language wd:Q150 .
Language: French
Total number of lexemes: 19746
```
This shows inconsistent behaviour.

### Expected Behavior
The command should validate the provided language and data type. If either does not exist, the system should gracefully return without executing the query and also suggest to the user what they can do to resolve it.

### Root Cause
The current implementation lacks validation checks for the existence of input languages and data types in the metadata files. Specifically, the language_metadata.json file plays a crucial role in this issue. It serves as the authoritative source for valid languages and their corresponding QIDs. When a user inputs a language or data type that is not present in this file, the CLI does not recognize it as invalid and proceeds with the query. This oversight results in misleading output and a poor user experience

### Proposed Solution
- [x] Introduce a validation function that checks for the existence of the provided language and data type.
- [x] Modify the CLI code in total.py to include a user-friendly message for non-existent languages.
- [x] Ensure that the main script correctly handles the update and setting of metadata, as defined in main.py, to improve user experience.

### Related Issues
This issue is closely related to #295 as it has to do with CLI

### Contribution
I would love to work and collaborate on implementing this improvement.

DeleMike:
- Implemented checks for non-existent languages and data types in the total command.
- Added informative error messages guiding users to update or set their language metadata.
- Enhanced feedback for improved usability of the CLI.

### Contributor checklist

---

### Description

This PR enhances the Scribe-Data CLI by validating user input for languages and data types. If a non-existent language is specified, the user will receive a clear message informing them that the language does not exist, along with instructions on how to update their language_metadata.json file using:

```bash
scribe-data update --metadata
```

Alternatively, users can manually set the metadata with:

```bash
scribe-data set-metadata -lang [your_language] -qid [your_qid]
```

These enhancements aim to improve user experience and ensure accurate data retrieval in the Scribe-Data CLI.

### Related issue
This issue is closely related to #295 

- fixes #328


catreedle:
Hi @DeleMike,
Great job on this! It's impressive how you've come up with this solution.

Just a couple of things to consider:
- We might need to revisit total.py later, as there could be some changes related to [this PR](https://github.com/scribe-org/Scribe-Data/pull/311).
- Do we think it's safe to allow users to update the language_metadata? Specifically, does this handle cases where a user might try to update an existing language (e.g., English) with an incorrect QID?

Let me know your thoughts when you get a chance. :)

catreedle:
I was browsing through the issues and found one that might be related to this. https://github.com/scribe-org/Scribe-Data/issues/293

DeleMike:
> Hi @DeleMike, Great job on this! It's impressive how you've come up with this solution.
> 
> Just a couple of things to consider:
> 
> * We might need to revisit total.py later, as there could be some changes related to [this PR](https://github.com/scribe-org/Scribe-Data/pull/311).
> * Do we think it's safe to allow users to update the language_metadata? Specifically, does this handle cases where a user might try to update an existing language (e.g., English) with an incorrect QID?
> 
> Let me know your thoughts when you get a chance. :)


Yes, we might need to revisit it after it has been merged. Thanks @catreedle! 

concerning the ability to allow users (the developer) to update the language metadata is a good option IMHO. In Flutter and other systems, there's always a need to sometimes update your dependencies or refetch them(`flutter pub get`) just to make sure they are cached properly/available for use. However, I agree with you when you say, what if the user updates, say, "English" with the wrong QID, how do we tackle that? 

I will think of ways soon and drop them in this discussion.

DeleMike:
> I was browsing through the issues and found one that might be related to this. #293

hmm... did not see that. 
It is closely related, right ? @andrewtavis. What can we do about this?

One thing, this issue originated when the `Scribe-Data` was returning some kind of default values for languages it could not find in the `language_metadata.json` file. Hence, I came up with this initial fix which led to the procedure of updating the metadata

```python
 # when we want to run `scribe-data t -l Latin` and Latin doesn't exist in the metadata hence we cannot provide a QID for it
 language_qid = get_qid_by_input(language)
 data_type_qid = get_qid_by_input(data_type)
 
 if not language_qid:
        print(
            "The specified language does not exist. Please update your language_metadata.json file by using:\n"
            "`scribe-data update --metadata`\n"
            "Alternatively, you can manually set it with:\n"
            "`scribe-data set-metadata -lang [your_language] -qid [your_qid]`.\n\n"
            "This will ensure that you can fetch the correct data."
        )
        return
```

We will basically inform the user that the specified language does not exist and they should update the metadata file. We also terminate the whole process early with the return keyword so that we exit from the procedure gracefully after telling the user what to do.

Output of running `scribe-data t -l Latin` if it does not exist:
```bash
The specified language does not exist. Please update your language_metadata.json file by using:
`scribe-data update --metadata`
Alternatively, you can manually set it with:
`scribe-data set-metadata -lang [your_language] -qid [your_qid]`.

This will ensure that you can fetch the correct data.
```

andrewtavis:
Ok @DeleMike, I've had a bit of time to look this over. Big thing is that this functionality is useful now, but shouldn't be in the future as we shouldn't be releasing Scribe-Data or merging in language querying files if the language isn't in the metadata file. @OmarAI2003 will be expanding the language metadata file in #293, so then this should be fine 🤔 I agree with @catreedle that we don't necessarily want the user to have control over these internal files.

There is definitely some value here though 😊 Specifically maybe we can change this so that it does the following:
- A new file `/src/scribe_data/check_language_metadata.py` that would include some of the functionality you have here to check the `language_data_extraction` directory to see if the languages in the directory are included in the metadata file
- A GitHub workflow in [.github/workflows](https://github.com/scribe-org/Scribe-Data/tree/main/.github/workflows) that runs on PRs that would run `check_language_metadata.py` and throw an error if the `language_data_extraction` directory hasn't been updated

How does this sound?

src/scribe_data/cli/cli_utils.py
```diff
@@ -36,11 +38,19 @@
 )
 DATA_DIR = Path(DEFAULT_JSON_EXPORT_DIR)
 
-with LANGUAGE_METADATA_FILE.open("r", encoding="utf-8") as file:
-    language_metadata = json.load(file)
+try:
```
andrewtavis, line 41: These try-excepts are fine, but can we not set `language_metadata = {"languages": []}` at the end of them (and the same for data types). If these files are not being loaded in then we have serious problems. If there's not a test for it already - that both of them are read and are in a specific location - then we should write one!

CC @DeleMike and @catreedle for checking for this test :)

src/scribe_data/cli/cli_utils.py
```diff
@@ -52,6 +62,129 @@
 }
 
 
+def get_available_languages() -> list[tuple[str, str]]:
```
andrewtavis, line 65: Here we'd just return the languages for the `check_language_metadata.py` file :) There's a degree of hard coding here with the `query_verbs.sparql` that we should avoid.

Maybe there is some value with letting a Scribe-Data developer set the metadata file with a language and a QID of their choice though 🤔 They could also use the QID during testing, but then that has specific return functionality now that's getting all data types by default instead of what's in Scribe-Data already. This would allow us to do a simple command in the docs to update the metadata file, but we should make this clear that it's for development only. We don't want the end user to add a language-qid pair to the metadata as they wouldn't have the ability to write the queries (if say they install with PIP).

src/scribe_data/cli/cli_utils.py
```diff
@@ -52,6 +62,129 @@
...
+            sparql_file_path = lang_folder / "verbs" / "query_verbs.sparql"
+            qid = extract_qid_from_sparql(sparql_file_path)
+            if qid:
+                available_languages.append((lang_name, qid))
+            else:
+                available_languages.append((lang_name, ""))
+    return available_languages
+
+
+def extract_qid_from_sparql(file_path: Path) -> str:
```
andrewtavis, line 86: `extract_qid_from_sparql`, `check_and_update_languages` and `update_language_metadata` wouldn't be needed as this would be covered by the check workflow on PRs that will assure that the metadata file is appropriate :) I'm still on the fence when it comes to `set_metadata` though 🤔

src/scribe_data/cli/main.py
```diff
@@ -30,13 +30,16 @@
...
 from scribe_data.cli.version import get_version_message
+from scribe_data.cli.cli_utils import check_and_update_languages, set_metadata
 
 LIST_DESCRIPTION = "List languages, data types and combinations of each that Scribe-Data can be used for."
 GET_DESCRIPTION = (
     "Get data from Wikidata and other sources for the given languages and data types."
 )
 TOTAL_DESCRIPTION = "Check Wikidata for the total available data for the given languages and data types."
 CONVERT_DESCRIPTION = "Convert data returned by Scribe-Data to different file types."
+UPDATE_DESCRIPTION = "Update the metadata file with available languages and QIDs."
```
andrewtavis, line 41: Main would need the update functionality removed as it'll be covered by the new workflow on PRs :)

src/scribe_data/cli/total.py
```diff
@@ -206,6 +206,16 @@ def get_total_lexemes(language, data_type, doPrint=True):
     else:
         data_type_qid = get_qid_by_input(data_type)
 
+    if not language_qid:
+        print(
+            "The specified language does not exist. Please update your language_metadata.json file by using:\n"
```
andrewtavis, line 211: We'll need to check these outputs once the updates are done :)

src/scribe_data/cli/cli_utils.py
```diff
@@ -52,6 +62,129 @@
...
+            # Move the file pointer to the beginning and overwrite the existing file
+            file.seek(0)
+            json.dump(language_metadata, file, ensure_ascii=False, indent=4)
+            file.truncate()  # Remove any leftover data
+        print("Language metadata updated successfully.")
+    except (IOError, json.JSONDecodeError) as e:
+        print(f"Error updating language metadata: {e}")
+
+
+def set_metadata(language_name: str, qid: str):
```
andrewtavis, line 153: If we do keep `set_metadata`, then we should include an iso argument :)

andrewtavis: Also @DeleMike, it'd be great if for those functions that are kept that you check some of the other docstrings and use similar formatting. As the function docstrings are written now, they won't be parsed correctly for the readthedocs documentation :)

andrewtavis:
Specifically of interest to me now though, @DeleMike and @catreedle, and maybe something that you two would like to take on and maybe @DeleMike can take the lead: We could repurpose this code as well to have another check of the actual queries. This idea came from `extract_qid_from_sparql`. There obviously is a lot of copying and pasting going on, and I've found some errors in the PRs, but maybe some slipped though. We should also have a GitHub Actions workflow that will check all queries for the following:

- Is the language QID the one that it's supposed to be given the directory
- Is the data type the one it's supposed to be given the directory

We'd be able to use the language and data type metadata files for this. For the first test we're seeing if the first QID is the correct language via parsing `?lexeme dct:language wd:Q12345` (we'd need to continue to write like this, but it's fine). From there for the data type check we can just check to see if any of the QIDs for the other data types are in the query - i.e. "Hey this is an adjectives query. Why is the QID for nouns in here???".

Let me know how the above sounds! I think that the above serves to make use of the code here in intuitive ways so the efforts aren't wasted, and we get a great PR testing setup that checks the metadata files and queries to make sure that we're not introducing things that don't work 😊

catreedle:
> Specifically of interest to me now though, @DeleMike and @catreedle, and maybe something that you two would like to take on and maybe @DeleMike can take the lead: We could repurpose this code as well to have another check of the actual queries. This idea came from `extract_qid_from_sparql`. There obviously is a lot of copying and pasting going on, and I've found some errors in the PRs, but maybe some slipped though. We should also have a GitHub Actions workflow that will check all queries for the following:
> 
> * Is the language QID the one that it's supposed to be given the directory
> * Is the data type the one it's supposed to be given the directory
> 
> We'd be able to use the language and data type metadata files for this. For the first test we're seeing if the first QID is the correct language via parsing `?lexeme dct:language wd:Q12345` (we'd need to continue to write like this, but it's fine). From there for the data type check we can just check to see if any of the QIDs for the other data types are in the query - i.e. "Hey this is an adjectives query. Why is the QID for nouns in here???".
> 
> Let me know how the above sounds! I think that the above serves to make use of the code here in intuitive ways so the efforts aren't wasted, and we get a great PR testing setup that checks the metadata files and queries to make sure that we're not introducing things that don't work 😊

This is very thoughtful. I'm willing to work/collaborate on it. 😊

src/scribe_data/cli/cli_utils.py
andrewtavis: Given the reaction here, @catreedle, do you want to check the tests to see if we have some for those two files being accessible, and if not you can open a PR with tests for it? :)

catreedle: Sure, allow me some time as I’m still getting familiar with testing. I'll check and get back to you later. :)

andrewtavis: Sounds good!

catreedle: From what I see, there aren’t any specific tests to verify whether both language_metadata.json and data_type_metadata.json are being read, but the tests depend on these files. Deleting one causes an error during test runs. I've written tests to check the accessibility of these files, but they won’t run if the files are missing. I’m unsure about adding these tests and could use some help with this. :)

catreedle: I guess I could put it in a new test file for this if it's still needed?

DeleMike:
Hey, @andrewtavis and @catreedle thanks folks for the review😄. 
so helpful. I am going through everything as I type. 

I just wanted to drop my initial thoughts. From all I have read, we can generate a set of new issues from the works in this PR. I will be dropping my comments!

src/scribe_data/cli/cli_utils.py
DeleMike: Yes, I guess you could but wait for a final go-ahead :)

src/scribe_data/cli/cli_utils.py
DeleMike: You are right. I thought about it only returning language names but I felt we could do better or create like a "standard". i.e. We will have to ensure that we have `query_verbs.sparql` exists for all directories. And I've checked, the `French` directory has`query_verbs_1.sparql` and `query_verbs_2sparql`, so I guess it won't be helpful.

src/scribe_data/cli/cli_utils.py
andrewtavis: We can have two minor tests just like you described, @catreedle :) Feel free to put them in an appropriate directory/file that's for the whole application. I'll move it if needed 😊

src/scribe_data/cli/cli_utils.py
DeleMike: Yeah, I agree with why we should move them. Great point 😄

I thought that the `set_metadata` would be useful if, for some reason, we could not automatically update the metadata(that is, when it was still in use), then we can call `scribe-data set-metadata -lang French -qid Q150` for example, and this updates our metadata file.

src/scribe_data/cli/total.py
DeleMike: Yeah, sure! Thanks @andrewtavis !

src/scribe_data/cli/cli_utils.py
DeleMike: oh! I did not know about this. will update :)

andrewtavis:
I think we should definitely get this down to a point where we can merge and use the code from it as example codes within the issues we make 😊 Let me know if I should make the issues and tag you both!

DeleMike:
> Specifically of interest to me now though, @DeleMike and @catreedle, and maybe something that you two would like to take on and maybe @DeleMike can take the lead: We could repurpose this code as well to have another check of the actual queries. This idea came from `extract_qid_from_sparql`. There obviously is a lot of copying and pasting going on, and I've found some errors in the PRs, but maybe some slipped though. We should also have a GitHub Actions workflow that will check all queries for the following:
> 
> * Is the language QID the one that it's supposed to be given the directory
> * Is the data type the one it's supposed to be given the directory
> 
> We'd be able to use the language and data type metadata files for this. For the first test we're seeing if the first QID is the correct language via parsing `?lexeme dct:language wd:Q12345` (we'd need to continue to write like this, but it's fine). From there for the data type check we can just check to see if any of the QIDs for the other data types are in the query - i.e. "Hey this is an adjectives query. Why is the QID for nouns in here???".
> 
> Let me know how the above sounds! I think that the above serves to make use of the code here in intuitive ways so the efforts aren't wasted, and we get a great PR testing setup that checks the metadata files and queries to make sure that we're not introducing things that don't work 😊

Lovely! I would love to improve this code.

From the quote, I understand that:

1. We want to repurpose existing code, particularly from `extract_qid_from_sparql`, to check the accuracy of queries.
2. We want **Language QID Check** and **Data Type Check** in our workflows for validation.

DeleMike:
> I think we should definitely get this down to a point where we can merge and use the code from it as example codes within the issues we make 😊 Let me know if I should make the issues and tag you both!

yes, please. we would love to work on the issues created! :)

andrewtavis:
Ok, let me know how you want to proceed here, @DeleMike :) I'm happy to clean this PR up with the code we want for now, and then also make the issues at the same time using the current functions as snippets. Let me know if you'd like to do any changes before - so I'll wait for a confirmation here and then finalize this one 😊

DeleMike:
Yeah, true!
For now, I'd rather you help us point in the direction we should go. That is, if there's any extra addition I'd make at this moment, it would all stem from your suggestions hence I would be happy if you could help adjust the PR now, and then make issues from it. The main reason is that we don't create a future complication. How about this?

or is there something I could quickly do so that it does not affect our codebase? the docstrings format issue?

andrewtavis:
I can get to the docstrings really quick, @DeleMike :) No stress at all. Just wanted to note it for the next time!

andrewtavis:
I'll do a quick fix of this one soon then and then make issues for the workflows 🚀

DeleMike:
> I can get to the docstrings really quick, @DeleMike :) No stress at all. Just wanted to note it for the next time!

oh okay! noted 😄

DeleMike:
> I'll do a quick fix of this one soon then and then make issues for the workflows 🚀

I'll be waiting for them :)

Thanks @andrewtavis for the support!

and thank you @catreedle. Let's get ready!! 💪😄

andrewtavis:
#339 and #340 have been made and I pinged you each in them :) Let me know if there are any questions!

DeleMike:
alright, checking...

andrewtavis (approved):
I decided against doing set metadata as well, @DeleMike, as I think we need to focus on making sure that the CLI options are for the end users and not for the development team. Thanks so much for these suggestions though! I think that the workflows are going to improve the project so much, and all of this was the catalyst 😊

DeleMike:
> I decided against doing set metadata as well, @DeleMike, as I think we need to focus on making sure that the CLI options are for the end users and not for the development team. Thanks so much for these suggestions though! I think that the workflows are going to improve the project so much, and all of this was the catalyst 😊

That's okay :-) 
And you are welcome!

src/scribe_data/cli/cli_utils.py
catreedle: added a PR [here](https://github.com/scribe-org/Scribe-Data/pull/343).  Looking forward to the review. :)

PR #334, opened 2024-10-13
fix interactive cli command

axif0:
### Contributor checklist

---

### Description
Fixed `skipping` message If user press `Overwrite`: No. 
Used `tqdm` for process bar. I feel like if we use another process bar from rich library it looks odd.
Move the Configuration Summary in middle.
Added log as `✔ Exported {language} {data-type} data.`,
Added failed to export check message if there is an error for exporting language and data-type.
Added `MAGENTA` color in process bar. Available colours are `[hex (#00ff00), BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE]`.
Also, in cli the `Get All` is not working at all. Can I work with it? @andrewtavis ?

image

### DIFF
--- a/src/scribe_data/cli/get.py
+++ b/src/scribe_data/cli/get.py
@@ -62,14 +62,14 @@ def get_data(
         outputs_per_entry : str
             How many outputs should be generated per data entry.
 
-        overwrite : bool
-            Whether to overwrite existing files (default: False).
+        overwrite : bool (default: False)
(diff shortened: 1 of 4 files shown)


andrewtavis:
> Also, in cli the `Get All` is not working at all. Can I work with it? @andrewtavis ?

Feel free to open an issue and start working on it, @axif0! :)

Issue #307, opened 2024-10-10
Incorrect Results for Non-existent Language/Data Type in total Command


catreedle:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

When a non-existent language or data type is provided to the total command, the system returns an incorrect number of lexemes instead of a "Not found" message.

**Steps to Reproduce**:
1. Run the total command with an invalid language or data type.
2. Observe that the command returns a number instead of indicating that the language or data type doesn't exist.

![bug](https://github.com/user-attachments/assets/88bb9e6e-dc58-4d35-a45b-3cd87cb47cdb)

**Expected Behavior**:
When a non-existent language or data type is provided, the command should return a clear message indicating "Total number of lexemes: Not found" instead of returning incorrect results.

**Proposed Solution**:
- [ ] Add a check function to validate if the provided language and data type exist in the [language_metadata](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/language_metadata.json) or [data_type_metadata](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/data_type_metadata.json).
- [ ] If either the language or data type does not exist, the system should return a "Not found" message without executing the query.

This issue is closely related to #295
This issue is blocked by #295

**Contribution**
I would love to work and collaborate on this.

catreedle:
Hi @andrewtavis, can you please add the blocked label and assign this to me? Thank you. :)

DeleMike:
hey @catreedle! could we collaborate on this issue? we could discuss it over Matrix. what do you think?

catreedle:
Hi @DeleMike. Sure thing, I’d be happy to collaborate on this. Just a heads up that it's currently blocked by another issue, but we can still discuss it. :)

DeleMike:
cool! I'll DM you on Matrix :)

andrewtavis:
Closed via #337, with the follow up work in #341 :) Thanks @catreedle!

PR #337, opened 2024-10-13
add validate language and data type #307

catreedle:
### Contributor checklist

---

### Description

This PR introduces a function to validate the language and data type. It raises an error to prevent get_total_lexemes from executing the query if either QID is None.

Tested this command

- `scribe-data total -l invalid_language -dt invalid_type`
- `scribe-data total -l english -dt invalid_type`
- `scribe-data total -l invalid_language -dt nouns`

All results `Total number of lexemes: Not found`.

Added some unit tests as well.
Feedbacks is appreciated. :)
### Related issue

- #307

### DIFF
--- a/src/scribe_data/cli/total.py
+++ b/src/scribe_data/cli/total.py
@@ -217,7 +217,7 @@ def get_total_lexemes(language, data_type, doPrint=True):
         validate_language_and_data_type(language_qid, data_type_qid)
 
     except ValueError as e:
-        print(e)
+        print(str(e))
         return
 
     query_template = """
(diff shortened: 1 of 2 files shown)


src/scribe_data/cli/total.py
```diff
@@ -212,6 +212,12 @@ def get_total_lexemes(language, data_type, doPrint=True):
     else:
         data_type_qid = get_qid_by_input(data_type)
 
+    # Validate language and data type, raise error if invalid
+    try:
+        validate_language_and_data_type(language_qid, data_type_qid)
+    except ValueError as e:
+        print(str(e))
```
andrewtavis, line 219: No need to call string within print :)

andrewtavis: Or not given the error :) I guess it does make sense that if it's an error then we need to convert it to its string.

andrewtavis (approved):
Thanks for this, @catreedle! We'll be using this in #341 as well 😊 I'll write to you over there to get some feedback on it :)

Issue #227, opened 2024-10-03
Expand Estonian data queries


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would look into expanding the [src/scribe_data/language_data_extraction/Estonian](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/language_data_extraction/Estonian) files with as much data as are possible from the current data on Wikidata. We can use code for getting data from other languages, and from there we can check Estonian data on Wikidata for what conjugations are available. We can then expand the query with optional selections of certain forms as is done in other SPARQL queries. The query can be tried on the [Wikidata Query Service UI](https://query.wikidata.org/) during development :)

Data types to include:
- [x] Nouns
- [x] Verbs
- [x] Adjectives
- [x] Adverbs
- [x] Prepositions
- [x] Emoji keywords

### Contribution

Happy to support via answering questions and also reviewing code once a PR has been made 😊

Ekikereabasi-Nk:
Hi @andrewtavis I'm interested in this. I will also like the create an issue for polish adjective extraction query

andrewtavis:
Just added a list of data types that we want to include to this issue :) Have marked those that are already done or have PRs open, and we can work on the others 😊 If the data type can't work, then we can move to the others and open up specific issues later :)

andrewtavis:
No stress, @Ekikereabasi-Nk!

andrewtavis:
Ok, @Ekikereabasi-Nk :) So, for the verbs query here I've mapped out sections that would be good to split things into. This query will likely also need to be split into other files, but my suggestion would be to write the whole thing and then split it. I can also assist more here, but if you'd like to work on the verbs query more you'd be welcome to!

andrewtavis:
Let's take what we've learned from the recent PRs and move on to verbs, @Ekikereabasi-Nk :) One thing, please pull the changes of main from time to time. I'm realizing that your branch is behind main, and it means that when I'm in your branch I can't check what the others have done.

Ekikereabasi-Nk:
Thank you @andrewtavis.

andrewtavis:
[...]

PR #345, opened 2024-10-14
 Estonian verb data query

Ekikereabasi-Nk:
### Contributor checklist

---

### Description

This pull request introduces a set of SPARQL queries that retrieve various Estonian verb forms

I tested the queries in the Wikidata query service

- #227


Ekikereabasi-Nk:
Hi @andrewtavis i done with the verb queries. It much. I"m done too with the emoji_keyword_generation, only had errors when trying to test of it. I will raise a PR soon

andrewtavis:
The queries are coming out a lot nicer, @Ekikereabasi-Nk :) One thing to note is that we really want one row per result. I think that we need so many files in this case because we're getting lots of results duplicated. There should also be no need for `SELECT DISTINCT` in any of these. Can you go back through and add in indicative and conditional to the queries for present tense and try to put singular and plural into one file? Once that's done and you're getting one row per Wikidata item, you can do the same for other cases?

Ekikereabasi-Nk:
Alright @andrewtavis will do so. Thank for the feedback

Ekikereabasi-Nk:
Hi @andrewtavis, I need your assistance. I need a clearer picture to understand "_One thing to note is that we really want one row per result_". While learning I came across the SAMPLE keyword commonly used in queries where there may be multiple results for a variable, and you want to reduce those results to one (it fit into the purpose of one row per result and avoid duplicate). But the downside of SAMPLE is selects only one value from the set of possible values, especially in the case of Estonian verb that has multiple valid forms. Should I use it or stick to other keyword like GROUP BY(which has complexity and query error "Unknown error: java.lang.StackOverflowError" . Thank you

Ekikereabasi-Nk:
Hi @andrewtavis I done with the verbs changes. Can I also be assign to python based issues, I'm open to joining a team working on an issue. Thank you

andrewtavis:
There's been a lot of progress here, @Ekikereabasi-Nk :) Big thing is that we need to make sure that we have all of the properties on each of the forms, and also that we're getting all of the forms. I'm not seeing third person forms, for instance. Do you want to take a further look into this?

Ekikereabasi-Nk:
Aright @andrewtavis  I will look at it. Thanks for the feedback

andrewtavis:
Feel free to work on the Python for #359 before you jump back over here, @Ekikereabasi-Nk. I know that you want to do Python, so as you wish, but you can take a break from this for now :)

andrewtavis (approved):
I'll bring this in and make the needed edits, @Ekikereabasi-Nk :) Thanks for the contribution!

PR #366, opened 2024-10-15
Create query_adverbs.sparql

Otom-obhazi:
adverb for chinese/mandarin

### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER


andrewtavis (approved):
This one needed a language filter, which you can see in the final version of the query, @Otom-obhazi :) Sometimes Chinese has multiple dialects on a single item, so we need to make sure that only one dialect is returned.

Issue #358, opened 2024-10-14
Allow for lists of arguments to be validated by the CLI validation functionality


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The function `validate_language_and_data_type` was recently added to the CLI utils [here](https://github.com/scribe-org/Scribe-Data/blob/7882f8e07d96b4d93ea0f06fd722fc109588e46f/src/scribe_data/cli/cli_utils.py#L158). One thing that it does is it skips lists of arguments as it's not set up to check them now. It'd be great if we could change that such that lists passed to CLI arguments like `--language` and `--data-type` could also be validated.

With this we'd also need testing:
- Testing to make sure that this list functionality works
- Testing of the ability to pass QIDs to the `validate_language_and_data_type` function

### Contribution

Happy to support here and review a PR when one's open 😊

axif0:
want to work on that.. @andrewtavis

andrewtavis:
Sure thing, @axif0 :) After this, it would be good to get your help finalizing some of the queries as this is a part of the project you haven't had much experience with yet 😊

andrewtavis:
Closed by #375 :) We can do minor issues for any further functionality that's needed, @axif0! Thanks for this 😊

PR #375, opened 2024-10-15
fix  lists of arguments to be validated

axif0:
### Contributor checklist

---

### Description
Added Testing to make sure that this list functionality works also ability to pass QIDs to the `validate_language_and_data_type` function.
Fixed- 
image

@andrewtavis  You might want to change the error message. Created a `except ValueError as e:` in `main.py` to make it looks good.

image

Also sure is this cover all from #358 as we wanted or not. 

### Related issue

- #358

### DIFF
--- a/src/scribe_data/cli/cli_utils.py
+++ b/src/scribe_data/cli/cli_utils.py
@@ -23,7 +23,7 @@ Utility functions for the Scribe-Data CLI.
 import difflib
 import json
 from pathlib import Path
-from typing import Union
+from typing import Union, List
 
 from scribe_data.utils import DEFAULT_JSON_EXPORT_DIR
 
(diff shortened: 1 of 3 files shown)


andrewtavis (approved):
Really great stuff, @axif0! We can experiment with this a bit more in the future. I set it up so that we get new lines so that it's a bit more clear rather than stringing it together with ands.

I also removed one of the tests as it seemed really similar to the following one, and there's actually a chance that the test might fail if we ever add support for Klingon 😅

```sql
# tool: scribe-data
# All Klingon (Q10134) nouns.
# Enter this query at https://query.wikidata.org/.

SELECT
  ?lexeme
  (REPLACE(STR(?lexeme), "http://www.wikidata.org/entity/", "") AS ?lexemeID)
  ?noun

WHERE {
  VALUES ?nounTypes {wd:Q1084 wd:Q147276} # Nouns and proper nouns

  ?lexeme dct:language wd:Q10134 ;
    wikibase:lexicalCategory ?nounTypes ;
    wikibase:lemma ?noun .
    FILTER(LANG(?noun) = "tlh-latn") .
}
```

axif0:
Oh.. I see Klingon in a test list_data_types [here](https://github.com/scribe-org/Scribe-Data/blob/ed6cfe793866124931ab673c563599d91796db85/tests/cli/test_list.py#L98). Although I should have give made up name. 😅

andrewtavis:
Ah ya I remember I talked about this with the original contributor there as well :) Let's leave it for now, and we can change it if we every do add Klingon support 😊

PR #398, opened 2024-10-17
Add Finnish verbs query

Ebeleokolo:
Hello @andrewtavis , This a new PR where I added new Finnish verbs query just by itself. 

Kindly review. Thank you.

### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER


andrewtavis (approved):
Quick fix to this, @Ebeleokolo :) I removed most of the returns so we know that this needs to be looked at when we have more data available :)

PR #406, opened 2024-10-17
Set PythonPath and fixed emoji generation for windows

SethiShreya:
Signed off by Shreya Sethi
shreyasethi47@gmail.com

- Update the docs to set the python path globally
- Fixed minor issues related to emoji generation for windows

Note: I haven't fixed file path issue in generate_emoji_keywords.py files as it is being worked on by others. Everything else is working fine

Run this command to test
```bash
scribe-data get --lang english -dt emoji_keywords
```

PR #407, opened 2024-10-17
Fixed pyICU capabilities for emoji functionalities

axif0:
### Contributor checklist

---

### Description
Fetched and install correct version pyICU from [here](https://github.com/cgohlke/pyicu-build/releases). 

`scribe-data get --lang german -dt emoji_keywords
`

By calling the command or if we call `emoji-keywords` from `interactive mode` we check if pyICU installed or not. If installed then we skip installation process and if not install then the installation process will begun.

We can add a question like `We need to install pyICU for emoji, As not installed do you want to install it?` like that. Or we can call it automatically from `requirement.txt` (probably)

image
 

From my side Works both ubuntu & windows perfectly.  😅

My hypothesis->
Well, when we want to install pyICU from pip it gets only compatible version for ubuntu. As pyICU has different releases, it doesn't detect first which releases it needs to install. like arm, amd  or cp310/311/312 or else. So by calling the api and matching our architecture like windows, we get our releases. And then pyICU doesn't cause any error.


axif0:
CC: @mhmohona  can you please review the pr if there is no problem?

andrewtavis:
Would you be able to fix the merge conflict here, @axif0? :)

src/scribe_data/check/check_pyICU.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    Check if PyICU is installed, and if not, download and install it dynamically.
+    If the OS is not Windows, directly use `pip install PyICU`.
+    """
+    package_name = "PyICU"
+
+    # Check if PyICU is already installed using pkg_resources
+    installed_packages = {pkg.key for pkg in pkg_resources.working_set}
+    if package_name.lower() not in installed_packages:
+        print(f"{package_name} not found. Installing...")
```
mhmohona, line 116: As a user I would be alarmed if suddenly a package gets auto installed. I think it would be nice to describe in one sentence why installing `PyICU` is needed. WHat do you think @andrewtavis?

mhmohona (changes requested):
I ran this branch in both Ubuntu and Windows, and tested by creating new environment, and it runs exactly like @andrewtavis gave the requirement to follow! Much appreciated @axif0. This PR is gonna be life saver. 

However, the changes in `get.py`, `process_uicode.py`, and `query_data.py` file - it is showing the whole file has been been changed, but when I went through the code, i found not whole not the whole file, rather partial part of it's been changed. It would be easier to track what changes has been made if the previous format was kept. 
What do you think @andrewtavis?

axif0:
@mhmohona I dunno ether, but accidentally I worked on anther branch, and had to copy the file and paste this branch. Sorry for the inconvenient. And thank you for the testing..

andrewtavis:
Let's maybe open a new version of this that has just the needed changes to make this work given changes here and in #406 :)

andrewtavis:
Closing in favor of #480 :) Thanks, @axif0!

PR #429, opened 2024-10-18
Expand Norwegian/Bokmal verb query

VNW22:
### Contributor checklist

---

### Description

This query expands the norwegian/Bokmal verbs query including all the forms. The queries have been tested at [query.wikidata.org](https://query.wikidata.org/)

### Related issue

- #ISSUE_NUMBER


andrewtavis:
Can we get the properties expanded out a bit here, @VNW22? I think that the data needs to be modeled a bit better by the Wikidata community as some forms have the exact properties but are different. With that being said, expanding the properties to the fullest will prepare us for when that remodel happens.

VNW22:
Absolutely, that makes sense. I’ll work on expanding the properties as much as possible to ensure we’re capturing all potential forms.

Issue #357, opened 2024-10-14
Add workflow to check project structure


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Along with #339 and #340 another check that's needed for the project is one that validates that some directories are being created in an appropriate way. Specifically within the [src/scribe_data/language_data_extraction](https://github.com/scribe-org/Scribe-Data/tree/main/src/scribe_data/language_data_extraction) we need all sub directories of languages to be consistently named, where as of now we have some directories that are different like noun instead of nouns. Ideally this would take the form of a workflow `check_project_structure.yaml` within [.github/workflows](https://github.com/scribe-org/Scribe-Data/tree/main/.github/workflows) that would call a script `check_project_structure.py` that would make sure that all directories under languages are data types of the project.

### Contribution

Happy to support with this and review when a PR is ready 😊

axif0:
want to assign.. if there is no problem.. @andrewtavis I'm learning github action thanks to you 😅 .

andrewtavis:
Sure thing, @axif0 :) Check #350 for another one and my comment for how to make it run on all commits to a PR 😊

andrewtavis:
Another idea for a workflow, @axif0: How about something that tells us which queries we're missing? So goes through and checks all languages and makes sure there's a directory for all the data types? Might fail for a quick moment, but that would assure that we have good support in the package 😊

axif0:
@andrewtavis Ok. what I understand is we need to create an workflow action that shows annotations (warning type) for missing queries. This workflow go through all the language folders and  sub_language, check if all data_types or not, if not append the warning. Also if folder exist, then check if query file exist or not.  

 [reference](https://stackoverflow.com/questions/74907704/is-there-a-github-actions-warning-state)

Am I right ? can you please elaborate a bit if you don't mind?

andrewtavis:
Roughly what you're describing, so make sure that all the directories are there and that in each of them is the query file. With that being said, we don't need this to be a warning. Let's let it fail and then that forces us to write the missing queries :)

andrewtavis:
[...]

PR #432, opened 2024-10-18
add missing file check function

axif0:
### Contributor checklist

---

### Description
Added `check_missing_query_files` function in `check_project_structure.py` github action to check all missing `data-type` query for all available `language`.

### Related issue

- #357


axif0:
@andrewtavis we have these missing query files to add.
image

src/scribe_data/check/check_project_structure.py
```diff
@@ -73,7 +73,32 @@
 BASE_DIR = "../language_data_extraction"
 
 
-def check_data_type_folders(path, language, subdir, errors):
+def check_for_sparql_files(folder_path, data_type, language, subdir, missing_queries):
+    """
+    Check if a data-type folder contains at least one .sparql file.
+
+    Args:
```
andrewtavis, line 80: @axif0 Can you please remember to format your docstrings in a way that's consistent with the rest of the project. As it's written this won't be parsed into the documentation.

andrewtavis (approved):
Ok this is really good, @axif0 :) Thanks for it. Please do note the docstrings in the future as with that this review would be pretty effortless. As you can see I commented the checks out for now as we're not close to being able to not have them commented, but we can get to it later 😊

src/scribe_data/check/check_project_structure.py
axif0: Sorry for this. And thank you. I will remember this in future.

PR #436, opened 2024-10-19
Nynorsk data

OmarAI2003:
### Contributor checklist

---

### Description

This PR will add files to extract adverbs, prepositions and adjectives using SPARQL. Currently, the Nynorsk directory supports extracting nouns, verbs, and this will add adverbs, adjectives and prepositions .


andrewtavis (approved):
Thanks for this, @OmarAI2003! Note that we don't need order by in queries as the order of the results doesn't matter, so it's just adding compute to the query :)

PR #441, opened 2024-10-19
Added Dagbani Nouns

GicharuElvis:
I Added Dagbani Nouns
Kindly have a look into this.


andrewtavis:
@GicharuElvis: There currently is no change in this PR, and the header is for nouns, but then the commits are referencing prepositions. Let's open up a new one here 😊

Issue #450, opened 2024-10-21
Create workflow to check query forms


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to document work that I've been doing to create a check for form identifiers within all Scribe-Data queries. The check goes through and derives the properties of all forms and create a new metadata file for lexeme forms from which a naming criteria is derived.

### Contribution

I'm working on this. Current status at time of writing is that I need to add in a check for whether all forms are actually returned in the query, that each returned form is unique, and that returned forms are ordered in the same way as the metadata file. From there also a minor check to see that the query docstring is correct would be good.

From there I need to go through and rename all the forms in the queries that have issues :)

DeleMike:
Hello @andrewtavis 👋🏾, well done for #450 !! 
I'm happy to assist you on this issue. Is there any task I can help with?

andrewtavis:
Thanks for your offer to help, @DeleMike :) Maybe I can solve the current issues with the queries later tonight, and from there we can make some issues for the extra features?

DeleMike:
Sure! I will ping you later, then!

andrewtavis:
Ok @DeleMike, with #451 in we now need the following:

- [] Add in a check for whether all forms are actually returned in the query (i.e. are there optional selections that are not brought up to the select statement)
- [] Add a check that each returned form is unique (never ?plural ?plural)
- [] Add a check that the returned forms are ordered in the same way as the metadata file and that their optional statements match (this is a simple as checking the top against the labels for the generated form labels)
- [] General testing and error handling

CC @axif0 who's also looking for Python work to do :) Do you all want to plan out the work here?

DeleMike:
Thanks @andrewtavis! 

We will look into it. @axif0 how about we break this into parts? 

---
**Some questions:**
For these checks, how do we wanna verify outputs? Are we gonna make network calls for all the checks? 

Also, I believe these checks will be part of our local tests, yes?

andrewtavis:
Honestly let's keep this to expanding the `lexeme_form_metadata.json` file. Yes it's more work for us, but doing network calls for all of this each time a PR commit is made would be way to much usage of a common database :) 

[...]

PR #476, opened 2024-10-23
Workflow query forms

axif0:
### Contributor checklist

---

### Description

- [X]  Add in a check for whether all forms are actually returned in the query (i.e. are there optional selections that are not brought up to the select statement)
- [X] Add a check that each returned form is unique (never ?plural ?plural)

CC: @andrewtavis Not sure if is it how we wanted and also try to touch all corner cases. 

### Related issue

- #450


andrewtavis:
Thanks for this, @axif0! Can you bring in main that has the language directories under Wikidata? :)

andrewtavis (approved):
Really nice work, @axif0! Appreciate this a lot as it's going to improve our ability to maintain the queries so much 😊

src/scribe_data/check/check_query_forms.py
```diff
@@ -183,14 +198,251 @@ def return_correct_form_label(qids: list):
...
+            returned forms are properly defined.
+    """
+
+    # Check if query has any OPTIONAL blocks.
+    optional_pattern = r"OPTIONAL\s*\{"
+    has_optional_blocks = bool(re.search(optional_pattern, query_text))
+
+    if has_optional_blocks:
+        return ""  # skip check for queries with OPTIONAL blocks
+
```
OmarAI2003, line 315: I’m sorry, but in this part I don't understand why you guyes @andrewtavis @axif0 skipped queries that contain the `OPTIONAL` statement:

```python
# Check if query has any OPTIONAL blocks.
optional_pattern = r"OPTIONAL\s*\{"
has_optional_blocks = bool(re.search(optional_pattern, query_text))

if has_optional_blocks:
    return ""  # skip check for queries with OPTIONAL blocks
```

Almost all the queries that have been or will be written will include an `OPTIONAL` statement, and this is where many people (including myself) often do errors.

PR #485, opened 2024-10-25
Update check query forms

VNW22:
### Contributor checklist

---

### Description
I have added the checks for whitespaces before comma and non white spaces before period and semi colon.
 I noticed that in  the check correct label when generating QID  (return f"{qid_label} {', '.join(not_included_qids)} not included in lexeme_form.metadata.json") where elements are separated by {', '} , because there is no whitespace before the comma, it would print "not included in lexeme_form.metadata.json" eventhough the QID is in the Lexeme_form, when its just a panctuation error. 

### Related issue

- #ISSUE_NUMBER


andrewtavis (approved):
Simplified the regexes a bit, @VNW22, but aside from that all's great! Thanks so much for this 😊

PR #487, opened 2024-10-25
feat enhancement Interactive UI language & data-type 

axif0:
### Contributor checklist

---

### Description
Using existing `prompt_toolkit` to show `languages` and `data-types` in a better way. 

@andrewtavis  Not sure if you like the idea. if not we can skip the PR. 

interactive2

### DIFF
--- a/src/scribe_data/cli/interactive.py
+++ b/src/scribe_data/cli/interactive.py
@@ -23,6 +23,8 @@ Interactive mode functionality for the Scribe-Data CLI to allow users to select 
 import logging
 from pathlib import Path
 from typing import List
+from prompt_toolkit import prompt
+from prompt_toolkit.completion import WordCompleter
 
 import questionary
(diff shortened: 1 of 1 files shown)


andrewtavis:
This is definitely an interesting idea, @axif0! What would you say are the benefits of it? :) I'm definitely inclines to merge 😊

axif0:
Umm...There are lots of languages to show in interactive mode on a single column. Initially my idea was to show them in a table with multiple column. But it seems that current libraries are not capable for that to show and as well as select them. 

Therefore I think it is a good idea to use alternative ideas.

andrewtavis (approved):
Thanks @axif0 :) Didn't explain my question well, but I actually was wondering if there were other options, with specifically the table view coming to mind. I think that this is the best functionality we could conceivably do 😊 Really appreciate the initiative!

PR #481, opened 2024-10-24
Check select where labels order

OmarAI2003:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch
- [x] I have tested my code with the `pytest` command as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-Data/blob/main/CONTRIBUTING.md#testing)

---

### Description

**Added `check_forms_order` Function**  
   - Introduced a new function to validate the order of variables in the SELECT clause (excluding `lexeme` and `lexemeID`) against the order in the WHERE clause for SPARQL query files.
   - Implemented pattern matching to handle cases where variable forms are extracted from the WHERE clause.
   - Includes handling for label services providing additional data, ensuring consistency between SELECT and WHERE clause variable orders.

### Related issue

- #450

PR #492, opened 2024-10-27
refactor and modify query forms

axif0:
### Contributor checklist

---

### Description
what I understand is `check_forms_order` are taking the lemma of the lexeme in select statement. it checks `dt_pattern` and `forms_pattern` in `where clause`. if the `select_vars[0] !=where_vars[0]` then it only gives `the error of order of variables in the SELECT statement does not match their order in the query.` 

Basically `check_undefined_return_forms` and `check_defined_return_forms` are for checking Unreturned forms. If a form are present in `while clause` but not stated in select statement, it gives error like `Unreturned  forms {form-name}`.

We can marge the overall functionalities into one function at once. 

I have a confution in [lexeme_form_metadata.json](https://github.com/scribe-org/Scribe-Data/blob/40c41ab087a2b2af21cb22220b8b12939fbcc8d4/src/scribe_data/resources/lexeme_form_metadata.json), the levels are missing like if I add verbsTest in select statement  and also in where clause, it should check the metadata file for lebel if lemma is there or not, if not should rise error. Is it what we wanted? 

### Related issue

- #450 
- 
### Related PR

Check select where labels order [#481](https://github.com/scribe-org/Scribe-Data/pull/481)

### DIFF
--- a/src/scribe_data/check/check_query_forms.py
+++ b/src/scribe_data/check/check_query_forms.py
@@ -77,6 +77,52 @@ def extract_forms_from_sparql(file_path: Path) -> str:
     return None
 
 
+# MARK: Extract Label
+
+
+def extract_form_rep_label(form_text: str):
+    """
+    Extracts the representation label from an optional query form.
+
+    Parameters
+    ----------
+        form_text : str
+            The text that defines the form within the query.
+
+    Returns
+    -------
+        str
+            The label of the form representation.
+    """
+    onto_rep_pattern = r"ontolex:representation .* ;"
+    if line_match := re.search(pattern=onto_rep_pattern, string=form_text):
+        rep_label_pattern = r".*\?(.*);"
+        if label_match := re.search(pattern=rep_label_pattern, string=line_match[0]):
+            return label_match[1].strip()
+
+
+# MARK: Extract QIDs
+
+
+def extract_form_qids(form_text: str):
+    """
+    Extracts all QIDs from an optional query form.
+
+    Parameters
+    ----------
+        form_text : str
+            The text that defines the form within the query.
+
+    Returns
+    -------
+        list[str]
+            All QIDS that make up the form.
+    """
+    qids_pattern = r"wikibase:grammaticalFeature .+ \."
+    if match := re.search(pattern=qids_pattern, string=form_text):
+        return [q.split("wd:")[1].split(" .")[0] for q in match[0].split(", ")]
+
+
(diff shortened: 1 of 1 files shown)


OmarAI2003:
> ### Related PR
> Check select where labels order [#481](https://github.com/scribe-org/Scribe-Data/pull/481)

It's strange that GitHub didn't notify me about mentioning a PR I authored 🤯

axif0:
> There is no while clause in SPARQL
Sorry I mean where..
 
I created a marged both functions and specify which error occurs for which reason. Maybe its more reliable. @OmarAI2003 Can you please check the PR? if it is giving corrected checking?

OmarAI2003:
> I created a marged both functions and specify which error occurs for which reason. 

Yeah I can see.

>Can you please check the PR? if it is giving corrected checking?

While it's super hard since all the queries are already fixed, I will try 😅

src/scribe_data/check/check_query_forms.py
```diff
@@ -556,30 +390,10 @@ def check_query_forms() -> None:
...
-
-        # Check the order of variables in the WHERE and SELECT clauses.
-        select_where_labels_matching = check_forms_order(query_text)
-        if not select_where_labels_matching:
-            error_output += f"\n{index}. {query_file_str}:\n  - The order of variables in the SELECT statement does not match their order in the query.\n"
+        # Check the order of variables in the WHERE and SELECT clauses, and if all forms are defined and returned.
+        forms_order_and_definition_check = validate_query_forms(query_text)
+        if forms_order_and_definition_check:
+            error_output += f"\n{index}. {query_file_str}:\n  - {forms_order_and_definition_check}\n"
             index += 1
```
OmarAI2003, line 397: Great job on removing your three functions and integrating them into the other function! But, I believe the new function name is too general. one can't  distinguish it from the file's main function `check_query_forms` . What do you think, @andrewtavis?

OmarAI2003 (changes requested):

src/scribe_data/check/check_query_forms.py
andrewtavis: I'll check this a bit later and see if any renaming is needed based on how it's reading in full :) Thanks for the check @OmarAI2003 and for the work in integrating it all, @axif0!

andrewtavis (approved):
Thank you for the refactor here, @axif0! Much easier to understand what's going on now :)

PR #503, opened 2024-10-31
Adds a new function to validate the order of variables in SPARQL base…

OmarAI2003:
…d on our lexeme_forms_json file

### Contributor checklist

---

### Description

This PR introduces a new function, `check_forms_order`, which orders column names based on the lexeme_forms json fie. The  checks the variables returned against the ground truth file, and sorts them accordingly.

**Changes:**
- **New Function:** `sort_columns_by_reference` implemented to handle the sorting of column names, ensuring correct ordering based on predefined criteria.
- **Integration:** integrated The function with the main validation function of the file

The SELECT clause will be ordered hierarchically:
1. Data type will be the first variable (if it's there).
2. Variable length (shortest first).
3. Lexicographical order within each length group, based on a predefined JSON list.
4. Variables not in the JSON file like those returned by the query service will be added at the end.

**Note:**
The SPARQL queries diagnosed by this function have not been modified yet so checks will fail. Once the function is agreed upon I will fix and push

### Related issue

- #450


andrewtavis:
Hey @OmarAI2003 👋 Looking at this for a moment, it looks like the capitalization might be a bit off for the expected forms? It's saying that we should have a bunch of UpperCamelCase returned forms, but the forms are written in lowerCamelCase. Would you be able to take a look at this? :)

OmarAI2003:
Welcome back @andrewtavis, So I wrote the function specifically for ordering purposes and didn’t account for the exact naming conventions in the file. The message printed by the function is intended to guide the query writer to match the correct order, without focusing on case sensitivity, as that’s handled by other functions 
a
Do you like the suggesting returned forms in the message to be lowerCamelCase as they should be in the file?

OmarAI2003:
So, should this be the case in the final message?
a1

andrewtavis:
Yes this is looking better, @OmarAI2003. Thanks!

OmarAI2003:
Glad to hear this! 
When you are done reviewing this, just let me know so I can start updating the SPARQL files diagnosed by this function.

andrewtavis:
General idea here @OmarAI2003 is that once the two PRs you have up are in, that I'll then do the new formatting process :) Let me know if there's anything I can do to help!

OmarAI2003:
Hi @andrewtavis, I’ve fixed the queries. The final thing that will need to change is the MARKS.

andrewtavis:
Ok, I can do that, @OmarAI2003 :)

andrewtavis (approved):
The quality of the queries really is amazing now, @OmarAI2003 😊 Really thank you so much for all of the hard work to get us to this state! 🚀

Issue #504, opened 2024-11-08
Add PR check for Windows builds


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As discussed in the last dev sync, this issue would add a check that the project will build for Windows to the already existing Mac and Linux builds :)

### Contribution

CC @axif0, who had interest in working on this and suggested it in the first place 😊 Feel free to write in here and I'll assign!

andrewtavis:
Thanks, @axif0!

PR #505, opened 2024-11-10
Workflow check for Windows builds

axif0:
### Contributor checklist

---

### Description

In GitHub Actions, the default shell on Windows runners is PowerShell. However, activating a virtual environment in PowerShell can sometimes lead to execution policy issues. To simplify cross-platform compatibility, we can force the use of Bash on Windows runners, as GitHub Actions provides Git Bash on Windows.

Added condition for Homebrew setup which conditioned to run only on macOS.

### Related issue

- #504

### DIFF
--- a/.github/workflows/python_package_ci.yaml
+++ b/.github/workflows/python_package_ci.yaml
@@ -14,10 +14,15 @@ jobs:
         os:
           - macos-latest
           - ubuntu-latest
+          - windows-latest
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Fantastic, @axif0! So happy that Windows is getting the support that it needs these days :) Thanks so much for your efforts in these regards 😊

Issue #534, opened 2024-12-23
Version comparison fails due to inconsistent version string formats


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

The version comparison in [get_version_message()](https://github.com/scribe-org/Scribe-Data/blob/dd50d8dcbdbd624e84da9d5897f1ecd6c8adb88d/src/scribe_data/cli/version.py#L56) incorrectly reports available updates when local and latest versions are the same, due to mismatched string formats `("v" prefix and "Scribe-Data" prefix inconsistencies).`

![image](https://github.com/user-attachments/assets/6f59e741-2f28-4ccb-bde0-78681e818f36)

**Contribution**

Happy to review a PR when one's open and support with answers to any questions! 😊

andrewtavis:
Thanks for opening this, @axif0! Great to see that you're testing so much of the package :) :)

harikrishnatp:
I've identified where it's going wrong. In the [get_version_message()](https://github.com/scribe-org/Scribe-Data/blob/dd50d8dcbdbd624e84da9d5897f1ecd6c8adb88d/src/scribe_data/cli/version.py#L56) function, the value fetched from Github only has Scribe-Data without the `v` prefix while the value stored in local_version has Scribe-Data with a `v`. I could just standardise the version by removing `Scribe-Data` prefix from both for comparing and then adding v.

Also, the `pip scribe-data --upgrade` command is wrong, I fixed that as well

axif0:
Nice thought.. Do you want to be assign on the issue and work on this? @harikrishnatp

harikrishnatp:
@axif0 Yeah, sure

axif0:
Great @harikrishnatp ~

andrewtavis:
Closed by #535 🚀 Really appreciate the two great contributions to the project so far, @harikrishnatp 😊 Hope we can work with you more! Please let us know if you'd have interest in another issue here or in another Scribe project ✨

harikrishnatp:
Thank you for the kind words and the feedback! I'll be more careful about keeping file changes to their respective PRs! 😄

harikrishnatp:
[...]

PR #535, opened 2024-12-24
Fix version comparison bug in get_version_message() due to prefix inconsistencies

harikrishnatp:
### Contributor checklist

---

### Description

When you run the command scribe-data -v, even if you have the latest version, it incorrectly reports that an update is available. This issue arises due to mismatched values fetched from get_local_version and get_latest_version. The fix involves stripping unnecessary values and comparing only the relevant ones. Additionally, the command to update scribe-data was incorrect and has been corrected.

For testing, I added cases where the local and latest versions are either unknown or different.

### Related issue

- #534

### DIFF
--- a/src/scribe_data/cli/version.py
+++ b/src/scribe_data/cli/version.py
@@ -44,19 +44,22 @@ def get_latest_version():
 
 
 def get_version_message():
-    local_version = f"Scribe-Data v{get_local_version()}"
+    local_version = get_local_version()
     latest_version = get_latest_version()
 
-    if (
-        local_version == "Unknown (Not installed via pip)"
-        or latest_version == "Unknown (Unable to fetch version)"
-    ):
-        return f"{local_version}"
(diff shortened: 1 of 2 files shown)


andrewtavis:
Looks like the file changes for #533 were also included in here, @harikrishnatp :) Just a quick note that we should try to avoid this in the future 😊

andrewtavis:
I made a quick change so that we have `vX.Y.Z` in the tests just to make sure that there's no expectation that these will be maintained and updated as we have upgrades :)

andrewtavis (approved):
All's wonderful here, @harikrishnatp :) Thank you so much as well for writing the tests! Really appreciate you taking this on and delivering a sustainable solution 😊

harikrishnatp:
Thank you for the kind words and the feedback! I'll be more careful about keeping file changes isolated to their respective PRs! 😄

PR #540, opened 2025-01-02
Add isort for import sorting + EOF fixes

### LINKED ISSUE #539: Add import sorting to pre-commit hooks and pr_ci
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something I'm realizing is that we're currently not enforcing import ordering in Scribe-Data. One way to do this would be to use [isort](https://pycqa.github.io/isort/), but to stay within the [ruff](https://docs.astral.sh/ruff/) ecosystem we would ideally use their sorting methods. We'd like these checks added to [.pre-commit-config.yaml](https://github.com/scribe-org/Scribe-Data/blob/main/.pre-commit-config.yaml) such that they're enforced in commits for those that install the hooks as well as [pr_ci.yaml](https://github.com/scribe-org/Scribe-Data/blob/main/.github/workflows/pr_ci.yaml).

### Contribution

Happy to support as I can for this and review the changes when they're sent via a PR! 😊

Ktyby21:
### Contributor checklist

---

### Description

This pull request adds the `isort` hook to the `.pre-commit-config.yaml` to ensure all imports are consistently sorted according to the Black profile. I ran `pre-commit run --all-files` and verified that imports in several files were successfully reformatted. After that, I ran `pytest` to confirm that all tests still pass without any issues.

### Related issue

- Closes #539

### DIFF
--- a/.github/workflows/pr_ci.yaml
+++ b/.github/workflows/pr_ci.yaml
@@ -42,8 +42,7 @@ jobs:
           echo PATH=$PATH >> $GITHUB_ENV
 
       - name: Run ruff format - Formatting check
-        run: ruff check .
+        run: ruff check . --fix
 
-      - name: Run ruff - Linting and import sorting check
-        if: always()
+      - name: Run Ruff linting - Import sorting and linting
         run: ruff check .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,13 +7,6 @@ repos:
       - id: check-yaml
       # - id: check-added-large-files
 
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.0
-    hooks:
-      - id: ruff
-        args: [--fix, --exit-non-zero-on-fix]
-      - id: ruff-format
-
   - repo: https://github.com/tcort/markdown-link-check
     rev: v3.13.6
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,9 +25,10 @@ repos:
     hooks:
       - id: sphinx-lint
 
-  - repo: https://github.com/pycqa/isort
-    rev: 5.13.2
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.8.5
     hooks:
-      - id: isort
-        name: isort (python)
-        args: ["--profile=black"]
+      - id: ruff
+        args: [--fix]
+
+      - id: ruff-format
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,3 +24,10 @@ repos:
     rev: v1.0.0
     hooks:
       - id: sphinx-lint
+
+  - repo: https://github.com/pycqa/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+        name: isort (python)
+        args: ["--profile=black"]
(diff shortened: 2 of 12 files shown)


andrewtavis:
Hey @Ktyby21 👋 Thanks much for the contribution! Could we also add a check for this in `pr_ci.yaml`? We can't enforce that people use pre-commit, so it'd be good to check for this as well :) Beyond this, were you able to check if there's a way to do all of this with Ruff? I worry that maybe at some point there's a divergence of sorting and then we'll run into issues where the auto format from Ruff is then invalid to commit. Or does Ruff just use isort in the background?

Thanks again so much for the quick work! Looking forward to bringing this in :)

andrewtavis:
Specifically from another project I'm seeing we need to do `ruff format` also in the checks. We have `ruff check` twice accidentally :) The first instance of `ruff check` in `pr_ci.yaml` should be `ruff format`.

Ktyby21:
Hi @andrewtavis ,

Thank you for the detailed feedback and for reviewing my pull request!

Regarding your suggestions:

1. **Adding a check in `pr_ci.yaml`**:
   - I can add a step to run `isort --check --diff` in the CI workflow to enforce import sorting. Would you prefer to keep using `isort` alongside Ruff, or should we transition entirely to Ruff for both linting and import sorting?

2. **Ruff vs. isort for import sorting**:
   - I understand your concern about potential conflicts between Ruff and isort. If we decide to use Ruff for import sorting, I can configure it to handle the sorting to avoid any discrepancies. Alternatively, if you'd like to continue using `isort`, I can ensure that Ruff's settings do not interfere with it.

Please let me know which approach you'd prefer, and I'll make the necessary adjustments accordingly.

Thanks again for your guidance!

andrewtavis:
Thanks for the follow up questions, @Ktyby21! Answers to them:

1. I guess I'd prefer to keep the full process using Ruff if possible :) 
2. Let's use Ruff for the import sorting as well! 

Reason for this is that I'd like the solution here to be directly transferable to the other project I mentioned. You'd be welcome to send along the same changes as a PR to [activist-org/activist](https://github.com/activist-org/activist) with the CI changes going in `ci_backend.yaml` if you'd like! 😊

Let me know if there are further questions :)

Ktyby21:
Hi @andrewtavis,
I’ve pushed the updated branch with the latest ruff-format fixes. Please let me know if there are any additional steps needed.

andrewtavis (approved):
All's looking really great to me, @Ktyby21 :) Thanks for the great communication on this and for an amazing result! Looking forward to working with this tooling in the future!

andrewtavis:
Do you have an interest in doing a similar PR to [activist-org/activist](https://github.com/activist-org/activist), @Ktyby21? Also another thing I'm wondering about is maybe adding docstring validation using the [numpydoc pre-commit hook](https://numpydoc.readthedocs.io/en/latest/validation.html). Would you have interest in looking in to that? Big thing is that for here it would be great if we would not check the `./test` directory as I don't think we need to be writing full docstrings for all tests when the name of the test should normally document it, so it'd be great if we could figure out how to pass a directory path to the hook so it just checks `./src` 😊

PR #545, opened 2025-01-08
Introduced QID search in translation.

axif0:
### Contributor checklist

---

### Description
Added QID search for total, translation and forms. Even a language is not present in [language_metadata.json](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/language_metadata.json) we can use QID to get  total, translation and forms.
```
scribe-data get -l Q9217 -wdp scribe -od exported_json -a
```
image

### For total
```
scribe-data total -l Q9217 -wdp scribe
```
image

### DIFF
--- a/src/scribe_data/cli/total.py
+++ b/src/scribe_data/cli/total.py
@@ -24,7 +24,6 @@ from http.client import IncompleteRead
 from typing import List, Union
 from urllib.error import HTTPError
 
-import requests
 from SPARQLWrapper import JSON
 
 from scribe_data.utils import (
@@ -34,6 +33,7 @@ from scribe_data.utils import (
     language_metadata,
     language_to_qid,
     list_all_languages,
+    check_qid_is_language,
 )
 from scribe_data.wikidata.wikidata_utils import parse_wd_lexeme_dump, sparql
 
@@ -122,39 +122,6 @@ def get_datatype_list(language):
 
     else:  # return all data types
         return data_type_metadata
-
-
-def check_qid_is_language(qid: str):
-    """
-    Parameters
-    ----------
-    qid : str
-        The QID to check Wikidata to see if it's a language and return its English label.
-
-    Outputs
-    -------
-    str
-        The English label of the Wikidata language entity.
-
-    Raises
-    ------
-    ValueError
-        An invalid QID that's not a language has been passed.
-    """
-    api_endpoint = "https://www.wikidata.org/w/rest.php/wikibase/v0"
-    request_string = f"{api_endpoint}/entities/items/{qid}"
-
-    request = requests.get(request_string, timeout=5)
-    request_result = request.json()
-
-    if request_result["statements"]["P31"]:
-        instance_of_values = request_result["statements"]["P31"]
-        for val in instance_of_values:
-            if val["value"]["content"] == "Q34770":
-                print(f"{request_result['labels']['en']} ({qid}) is a language.\n")
-                return request_result["labels"]["en"]
-
-    raise ValueError("The passed Wikidata QID is not a language.")
 
 
 # MARK: Print
--- a/src/scribe_data/utils.py
+++ b/src/scribe_data/utils.py
@@ -23,7 +23,6 @@ Utility functions for data extraction, formatting and loading.
 
 import ast
 import contextlib
-import requests
 import json
 import os
 import re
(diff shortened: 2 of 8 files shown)


src/scribe_data/cli/total.py
```diff
@@ -124,39 +124,6 @@ def get_datatype_list(language):
         return data_type_metadata
 
 
-def check_qid_is_language(qid: str):
```
axif0, line 127: Had to move the `check_qid_is_language` method in utils.py for ImportError.

image

src/scribe_data/wiktionary/parse_dump.py
```diff
@@ -101,120 +102,19 @@ def _build_iso_mapping(self) -> dict:
...
-            # Build translations from sense glosses.
-            translations = {}
-            for sense in lexeme.get("senses", []):
-                for sense_lang_code, gloss in sense.get("glosses", {}).items():
-                    if sense_lang_code in self.iso_to_name:
-                        translations[sense_lang_code] = gloss["value"]
+        for language in self.target_iso:
+            if (
+                language.startswith("Q")
+                or language.startswith("q")
```
wkyoshida, line 108:
```suggestion
                language.lower().startswith("q")
```
Could do the above potentially? 

nit - just a suggestion really

src/scribe_data/utils.py
```diff
@@ -736,3 +737,70 @@ def check_index_exists(index_path: Path, overwrite_all: bool = False) -> bool:
...
+    api_endpoint = "https://www.wikidata.org/w/rest.php/wikibase/v0"
+    request_string = f"{api_endpoint}/entities/items/{qid}"
+
+    request = requests.get(request_string, timeout=5)
+    request_result = request.json()
+
+    if request_result["statements"]["P31"]:
+        instance_of_values = request_result["statements"]["P31"]
+        for val in instance_of_values:
+            if val["value"]["content"] == "Q34770":
```
wkyoshida, line 768: :thinking: Coming across this.. 

was actually wondering if we do have it documented somewhere what `P31` or `Q34770` refer to in Wikidata (and also all other important IDs that we reference in code too). 

If not, should we? Was thinking that a quick markdown table could suffice really.

src/scribe_data/utils.py
```diff
@@ -736,3 +737,70 @@ def check_index_exists(index_path: Path, overwrite_all: bool = False) -> bool:
...
+        An invalid QID that's not a language has been passed.
+    KeyError
+        The ISO code for the language is not available.
+    """
+
+    api_endpoint = f"https://www.wikidata.org/w/api.php?action=wbgetentities&ids={qid}&props=claims&format=json"
+    response = requests.get(api_endpoint)
+    data = response.json()
+    try:
+        return data["entities"][qid]["claims"]["P305"][0]["mainsnak"]["datavalue"][
```
wkyoshida, line 799: > ... was actually wondering if we do have it documented somewhere what P31 or Q34770 refer to in Wikidata (and also all other important IDs that we reference in code too) ...

ditto

wkyoshida (commented):
:smile:

src/scribe_data/utils.py
wkyoshida: Or actually - scratch that

It might be more helpful perhaps if that information is closer to where it's used in code. 
So in-source comments instead?

andrewtavis: Does another metadata file make sense for this? That way we reference an object and get the QID from a human readable object key?

wkyoshida: > Does another metadata file make sense for this?

hmm could be :thinking: 

Were you thinking of something like this?

```python
        instance_of_property = wikidata["property"]["instance-of"]
        instance_of_values = request_result["statements"][instance_of_property]
        for val in instance_of_values:
            if val["value"]["content"] == wikidata["entity"]["language"]:
```

andrewtavis: Along the lines of this, yes :)

andrewtavis (approved):
Thanks for the changes here, @axif0! This will make it much easier to figure out when we have enough translations on Wikidata to then make the switch of the translation functionality viable 😊

andrewtavis:
Quick note, @wkyoshida, that the above changes were made with @axif0 in the call and we discussed them while we prepared this for merging :) Just so you know that we did actively discuss the feedback and I didn't just do it myself 😇

PR #550, opened 2025-01-12
Add Workflow to Build and Push Docker Images in Docker Hub

mhmohona:
### Contributor checklist

---

### Description

This GitHub Actions workflow builds and pushes a Docker image to Docker Hub whenever there is a **push to the master branch** or a **manual trigger** using `workflow_dispatch`. It simplifies Docker image deployment and supports multi-platform builds.

#### Requirements:
- **Docker Hub Credentials**:
  - `DOCKERHUB_USERNAME`: Docker Hub username.
  - `DOCKERHUB_TOKEN`: Docker Hub personal access token.

### Related issue

- #150

### DIFF
--- a/.github/workflows/docker-image.yaml
+++ b/.github/workflows/docker-image.yaml
@@ -1,0 +1,33 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push the image
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          tags: asif570/testing-workflow:version0asif


mhmohona:
@wkyoshida mention about this workflow long ago, but I completely forgot to add it. 😅

.github/workflows/docker-image.yaml
```diff
@@ -0,0 +1,33 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    branches:
+      - master
```
axif0, line 6:
```suggestion
      - main
```

Aren't our default branch is main?

axif0:
Also can we add a checker if the repo is `scribe-org/Scribe-Data` ? like in [matrix_dev_sync_reminder](https://github.com/scribe-org/Organization/blob/main/.github/workflows/matrix_dev_sync_reminder.yml#L12)

andrewtavis:
Ya that'd be really great so forks don't try to push :)

.github/workflows/docker-image.yaml
andrewtavis: Beyond this, I think that it makes sense that the trigger for this is a release and not a push to `main` as we would only want full releases to go to Docker Hub.

andrewtavis:
> Ya that'd be really great so forks don't try to push :)

Beyond my comment above, let's make sure to add an explicit comment to the code that checks for the repo, as if someone does want to fork and host their fork on Docker Hub, then there's really no problem with that. It would be good if we made it clear where the change would need to go to disable the repo check 😊

andrewtavis:
Checking in on this here :) @mhmohona, would you be able to add in the check for which repo this is in as @axif0 mentioned [here](https://github.com/scribe-org/Scribe-Data/pull/550#issuecomment-2585762245)? We can then move towards finalizing this 😊

andrewtavis (approved):
Is looking great to me, @mhmohona! Thanks for the reactions to the review and improving the workflow 😊 I'll make sure that the credentials are in for the next release and then we'll see how this works! 🚀

Issue #573, opened 2025-02-26
Handling alternative forms in the downloaded data


fnielsen:
### Terms

- [x] I have searched [open and closed data issues](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Adata+)
- [x] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Languages

Danish

### Description

For some Danish lexemes there exists alternative forms. For instance, see imperative for the lexeme "determinere" at https://www.wikidata.org/wiki/Lexeme:L1010961

scribe-data yields
```
"L1010961": {
"passiveInfinitive": "determineres",
"verb": "determinere",
"imperative": "determinér",
"presentParticiple": "determinerende",
"pastParticiple": "determineret",
"passivePresent": "determineres",
"passivePreterite": "determineredes",
"activeInfinitive": "determinere",
"activePresent": "determinerer",
"activePreterite": "determinerede",
"lastModified": "2024-09-05T14:09:22Z"
},
```
Here there is only one imperative. I wonder if that is intended?

andrewtavis:
@fnielsen: Decision here is that we'll be comma separating these values in the same way that we have planned for genders and other forms in #564 :) We'll try to bring that in soon, and then the result of your command would then have the two forms for imperative in a comma separated string in alphabetical order. Your thoughts on this for a solution would be much appreciated!

andrewtavis:
Closed by #564 🚀 Thanks for integrating the work within the open PR to resolve this so quickly, @axif0 😊

Please let us know when you have a moment if the mentioned solution is sensible for you, @fnielsen :) Appreciate you opening these issues!

fnielsen:
Regarding comma: There might be idioms and the like that have commas.

andrewtavis:
Very true :) I was thinking about this as well 🤔 Can you think of another separator that might make sense? As of now we're mostly getting forms back, but maybe something like a bar (`|`) would be a good separator that's simple, but makes it clear that that it's likely been added by Scribe-Data? I don't think that a semicolon would indicate this as much :)

fnielsen:
The pipe may be ok. I tried
```
SELECT  
  ?lexeme
WHERE {
  ?lexeme ontolex:lexicalForm / ontolex:representation ?representation .
  FILTER (CONTAINS(?representation, "|"))
}
LIMIT 1
[...]

PR #577, opened 2025-03-04
Added pipe-separated values along with lexeme dump  

axif0:
### Contributor checklist

---

### Description
- Changed the separator from comma (", ") to pipe ("| ") in both `parse_dump.py` and `format_data.py` for better consistency and readability when merging multiple form values including lexeme dump.
- Added detection of multiple forms using the pipe separator and added appropriate warning messages in both files when multiple forms are found and merged.

### Related issue

- #573

### DIFF
--- a/src/scribe_data/wikidata/format_data.py
+++ b/src/scribe_data/wikidata/format_data.py
@@ -6,6 +6,8 @@ Formats the data queried from Wikidata using query_verbs.sparql.
 import argparse
 import collections
 
+from rich import print as rprint
+
 from scribe_data.utils import (
     export_formatted_data,
     load_queried_data,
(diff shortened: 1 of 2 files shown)


andrewtavis:
Just checking at this point, @axif0: Outputs of a WSQS process and a dump parse process will be the same?

axif0:
> Just checking at this point, @axif0: Outputs of a WSQS process and a dump parse process will be the same?

Actually, outputs of WSQS process is from [format_data.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/wikidata/format_data.py) and the dump parse process is from [parse_dump.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/wikidata/parse_dump.py) that's why the separation logic is implemented.

axif0:
I have a question, 

image
Left side of the JSON file is from lexeme dump, and right side of the JSON file is from WDQS.

Language [Spanish](https://www.wikidata.org/wiki/Q1321), Lexical category [noun](https://www.wikidata.org/wiki/Q1084), For the word `tífico` which is `L1346515`, (details: https://www.wikidata.org/wiki/Lexeme:L1346515)

Lexeme dump currently gets the forms which are on the page. But there is a mismatch between lexeme dump and WDQS.
I'm not so sure which one is correct.

andrewtavis:
I think the fields are not the same as we've yet to run the process that's being finished #578. Once that's done the outputs should be the same, but also for this it looks like we need to order the output fields based on the metadata file and put `lastModified` as the first field? Do you want to add that in to thus, @axif0?

andrewtavis:
So the separator is `" | "` with a space before now :)

Issue #575, opened 2025-02-28
Separate generated query forms based on keywords and a limited number of forms per query


andrewtavis:
### Terms

- [x] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aissue+label%3Afeature)
- [x] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

It would be great if the query generation process within [src/scribe_data/check/check_missing_forms](https://github.com/scribe-org/Scribe-Data/tree/main/src/scribe_data/check/check_missing_forms) would have a limited number of forms that would be added to a query such that we would have a maximum of six forms per query. We could group the forms based on their identifiers - splitting them based on the first one, and if that one is more than six then we split on the next one, etc.

### Contribution

Happy to support with this and review when it's finished! 🚀

andrewtavis:
CC @axif0 :)

andrewtavis:
On a call with @axif0 and assigning this given our conversation :)

andrewtavis:
Closed by #578 🚀 We're close to being all done here! 😊 Great work, @axif0!

PR #578, opened 2025-03-04
Refactor query generation by splitting query files

axif0:
### Contributor checklist

---

### Description

The `split_group_by_identifier` function takes language data containing missing features and splits them into manageable groups of up to 6 forms per query. By doing so, 979 files generated ([testing pr](https://github.com/axif0/test-scribe/pull/53)). 

- First groups by primary identifier
- If any group is too large (>6), splits by secondary identifier
- Combines small groups when possible to minimize the number of queries
- Finally generates queries for each optimized group

### Related issue

- #575


andrewtavis:
Absolutely amazing, @axif0! Really looking forward to reviewing this and closing all of the issues 😊

In the meantime do you want to start on your blogpost for a career in open source and the last one that's the review of the project?

axif0:
> In the meantime do you want to start on your blogpost for a career in open source and the last one that's the review of the project?

Sure! on it. :rocket:

src/scribe_data/check/check_missing_forms/normalize_forms.py
```diff
@@ -72,3 +81,107 @@ def get_sort_key(sublist):
...
+                    for (
+                        second_identifier,
+                        second_features,
+                    ) in second_level_groups.items():
+                        # Split into chunks of 6
+                        for i in range(0, len(second_features), 6):
+                            chunk = second_features[i : i + 6]
+                            final_groups.append(chunk)
+
+            # Now combine small groups if possible to reduce query files.
```
andrewtavis, line 151: I think that maybe we shouldn't do this and just understand that we'll be getting a lot of queries? Or are we not worried about this as if the forms are there then they won't be generated? I'm just thinking in terms of say `singular` and `plural` for nouns, where already for German these can't be in the same query, so maybe it makes sense to split on all first identifiers as well?

Let me know what you think, @axif0! 😊

axif0: > think that maybe we shouldn't do this and just understand that we'll be getting a lot of queries?

Actually I think we might need it, Here is an example scenario, 
```
//total 5 query file genarated.
final_groups = [
    [['Q1234', 'A1']],  # Group 1: 1 feature
    [['Q2345', 'B1'], ['Q2345', 'B2']],  # Group 2: 2 features
    [['Q3456', 'C1'], ['Q3456', 'C2'], ['Q3456', 'C3']],  # Group 3: 3 features
    [['Q4567', 'D1'], ['Q4567', 'D2'], ['Q4567', 'D3'], ['Q4567', 'D4']],  # Group 4: 4 features
    [['Q5678', 'E1'], ['Q5678', 'E2']]  # Group 5: 2 features
]
```
```
// total 2 query file genarated. 
optimized_groups = [
    [['Q1234', 'A1'], ['Q2345', 'B1'], ['Q2345', 'B2'], ['Q3456', 'C1'], ['Q3456', 'C2'], ['Q3456', 'C3']],  # Combined group of size 6 → 1 query file
    [['Q4567', 'D1'], ['Q4567', 'D2'], ['Q4567', 'D3'], ['Q4567', 'D4'], ['Q5678', 'E1'], ['Q5678', 'E2']]  # Combined group of size 6 → 1 query file
]
```
Without optimized arabic - nouns:
image

With optimized arabic - nouns:
image

axif0: > Ok :) But the big thing then is if we do separate queries ourselves it won't be changes, right? So if we split the German noun forms into separate queries it's not going to try to combine them back?

Ya, that`s right.

andrewtavis (approved):
Really excited for this, @axif0 😊 Amazing work in implementation and thanks also for the great collaboration in planning out these features! 🚀

I'll reach out about a call to run all of this and check the results together :)

PR #583, opened 2025-03-11
(Part 1 of 4) Arabic - finnish Queries for all data-types

axif0:
### Contributor checklist

---

### Description
Some files needed to splits, as it creates `run time error` or other error. So, I spited those.

For quickly check the queries, in [language_metadata.json](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/language_metadata.json) we can temporally remove alll language excluding arabic to finish. And in interactive mode we can go All language and all data-type, and then select `Run get data request with WDQS`.
 
Need to modify in [test_list.py](https://github.com/scribe-org/Scribe-Data/pull/583/files#diff-6432646e477391e70294e6ea00b00a6e6d1e965a0fc26f35d9cf5410cf909171), in `test_list_data_types_all_languages` and `test_list_data_types_specific_language` for newly added data-types.


andrewtavis:
Nice, @axif0! Does this also include the changes to make sure that ordering is correct and the price tests are ran before the PR? Or would those be in another PR?

axif0:
Also, there are some forms in dump for `Mandarin Chinese (Q727694)`, But those are not matched in [lexeme_form_metadata.json](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/lexeme_form_metadata.json). Thats why there are no `Mandarin Chinese (Q727694)` in the PR. 
Mandarin Chinese Forms like - [Rarely](vscode-file://vscode-app/tmp/.mount_cursorCgs3Bh/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), [Family name](vscode-file://vscode-app/tmp/.mount_cursorCgs3Bh/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), [Free variation](vscode-file://vscode-app/tmp/.mount_cursorCgs3Bh/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) and etc

andrewtavis:
> Also, there are some forms in dump for `Mandarin Chinese (Q727694)`, But those are not matched in [lexeme_form_metadata.json](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/resources/lexeme_form_metadata.json). Thats why there are no `Mandarin Chinese (Q727694)` in the PR. 
> 
> Mandarin Chinese Forms like - [Rarely](vscode-file://vscode-app/tmp/.mount_cursorCgs3Bh/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), [Family name](vscode-file://vscode-app/tmp/.mount_cursorCgs3Bh/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), [Free variation](vscode-file://vscode-app/tmp/.mount_cursorCgs3Bh/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) and etc

We can get to these later :) Let's focus on the current tasks and then slowly move towards a release with all the minor checks that come with that 😊

andrewtavis:
Minor change above for the generation script and the queries that it generated, @axif0 :) What do we want to do about the original queries? If we have a queries that have `_1`, `_2`, etc, normally we'd get rid of the base one without a number. Do you want me to go through and do this? Could also be done by numbering the original query and renumbering the new ones :)

axif0:
> If we have a queries that have `_1`, `_2`, etc, normally we'd get rid of the base one without a number.

I'm not quite sure what you mean by that. Could you please elaborate a bit ?

andrewtavis:
Sure thing! There's two situations for the queries that we have:

- There's one query, in which case we don't have `_1` or any number at the end of it
- There's more than one query, which means there's `_1`, `_2`, etc at the end of the queries
    - We shouldn't have a query without numbers though
    - So as of now we have `query_adjectives_1`, etc for Arabic, but we also have `query_adjectives` (this is incorrect as they all should have numbers in this case)

Did we agree that we'd do numbers at the end of all queries from now on? I can't remember 🤔 So by this I mean we'd never have a query without a number, which might make everything more easy, right?

axif0:
> Did we agree that we'd do numbers at the end of all queries from now on? I can't remember 🤔 So by this I mean we'd never have a query without a number, which might make everything more easy, right?

Oh, I got it. my mistake. Will update the workflow in a PR and updated the names in this PR.

andrewtavis:
Thanks, @axif0! I'll hope to get this in this evening :) Let's move on to the next set 😊

andrewtavis (approved):
All's looking amazing here, @axif0! Great work on this whole process. So nice to finally be finalizing all of this 🚀

PR #584, opened 2025-03-11
Add query forms validation step to GitHub workflow

axif0:
### Contributor checklist

---

### Description
- Added a checker if there is `No missing features/forms found.`
- Also, By using `python check_query_forms.py` we can actually check if the generated queries are corrected or not in the workflow.

### DIFF
--- a/.github/workflows/check_and_update_missing_query_forms.yaml
+++ b/.github/workflows/check_and_update_missing_query_forms.yaml
@@ -80,23 +80,24 @@ jobs:
       - name: check generated query forms
         working-directory: ./src/scribe_data/check
         run: |
+          python check_query_identifiers.py
           python check_query_forms.py
 
       - name: Post-run status
         if: failure()
-        run: echo "Generated SPARQL query forms check failed. Please fix the reported errors."
+        run: echo "Generated SPARQL query forms or identifiers check failed. Please fix the reported errors."
 
       - name: Generate PR Body
         id: pr-body
         run: |
           # Run the pr_body.py script with the missing features data.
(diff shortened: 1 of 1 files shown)


.github/workflows/check_and_update_missing_query_forms.yaml
```diff
@@ -101,6 +111,7 @@ jobs:
       #     cat $GITHUB_OUTPUT
 
       - name: Create Pull Request
+        if: steps.pr-body.outputs.body == 'All query forms and identifiers are correct.'
```
andrewtavis, line 114: I switched this line right here, @axif0, but maybe I didn't understand it at first 🤔 Don't we want to make sure that all's correct if we send a new PR? I also removed the double negative in the echo as it was a bit confusing :) So not `No missing features/forms found.`, but rather `All query forms and identifiers are correct.`. Let me know on this, and thanks for getting to this!

axif0: Actually, When there are no missing forms found, the workflow creates an error like [this action runner](https://github.com/axif0/test-scribe/actions/runs/13762723971/job/38482199536). Thats why I put that message actually.

axif0: Sorry, I should have elaborate that sooner. 😢

andrewtavis: All good! I'll switch it back and merge it in :)

andrewtavis (approved):
Thanks for the quick reaction to the need for query generation testing, @axif0! Great working with you on this 😊😊

Issue #206, opened 2024-10-02
Autosuggestions Data-type isn't accessible from the CLI


axif0:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Currently in [get.py](https://github.com/scribe-org/Scribe-Data/blob/128d089d4adcc12265ca44233fec8ebddf0a830f/src/scribe_data/cli/get.py), there is no logic set for Autosuggestions. 

![image](https://github.com/user-attachments/assets/e7824ca2-c08a-4021-8490-31104c3af937)

As in [process_wiki.py](https://github.com/scribe-org/Scribe-Data/blob/128d089d4adcc12265ca44233fec8ebddf0a830f/src/scribe_data/wikipedia/process_wiki.py#L322) function called [gen_autosuggestions](https://github.com/scribe-org/Scribe-Data/blob/128d089d4adcc12265ca44233fec8ebddf0a830f/src/scribe_data/wikipedia/process_wiki.py#L322) already defined. Can we can that in [get.py](https://github.com/scribe-org/Scribe-Data/blob/128d089d4adcc12265ca44233fec8ebddf0a830f/src/scribe_data/cli/get.py) ? If yes, I want to work on this issue.

andrewtavis:
Yes, this would be great, @axif0! But let's finish the current issues first :)

andrewtavis:
Discussing this a bit with @axif0 right now. We don't really think that the way that we're doing autosuggestions is something that would be applicable to other communities. Also, the long term plan is that we would actually not be using Wikipedia based calculated autosuggestions for this, but rather include a small LLM model in the language packs that would provide context based autosuggestions. Because of this, the work for this issue could potentially be to simply remove autosuggestions as an option for the CLI, and we will then use the notebook as needed in the Scribe community.

CC @wkyoshida and @mhmohona

Collins-Webdev:
Hello @andrewtavis 👋🏼,
I am eager to work on this bug.
 Can you asign me to it please 🙏🏽?

Collins-Webdev:
Hello @andrewtavis 👋🏼,
[...]

PR #595, opened 2025-03-23
Cli autosuggestions

catreedle:
### Contributor checklist

---

### Description

This PR adds a new file `generate_autosuggestions.py` that extracts steps from  [gen_autosuggestions.ipynb](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/wikipedia/gen_autosuggestions.ipynb)

Can be tested with command
```bash
scribe-data g -l italian -dt autosuggestions
scribe-data g -l indonesian -dt autosuggestions
```
or from the interactive CLI

### Related issue

- #206


axif0:
It takes only 57 minutes and half to download and parse Autosuggestions. :sweat_smile: 
 
image


Will look into this later, nice work @catreedle   :rocket:

catreedle:
> It takes only 57 minutes and half to download and parse Autosuggestions. 😅
> 
> image
> 
> @andrewtavis do we need to add `last-modified` parameter also in autosuggestions?
> 
> image
> 
> Will look into this later, nice work @catreedle 🚀

oh I set file limit
```python
files = download_wiki(
        language=language,
        target_dir=target_dir,
        file_limit=1,  # Limiting for development purpose
        dump_id=dump_id,
        force_download=force_download,
    )
```

and num words
```python
gen_autosuggestions(
        text_corpus,
        language=language,
        num_words=100,  # Limiting for development purpose
        ignore_words=None,
        update_local_data=True,
        verbose=True,
    )
 ```
 to save time and internet quota. It should be much faster than that. or did you set those off to download the whole dump?

andrewtavis:
Nice work to also use the testing arguments like this, @catreedle!

> @andrewtavis do we need to add `last-modified` parameter also in autosuggestions?

I'd honestly say no. Yes we could include it, but this current functionality is kind of a stopgap until next year where the general idea that we've had for a next GSoC project is to integrate language models into the keyboard applications. From there we'd actually have machine learning based autosuggestions based on context and not just single word autosuggestions, so at that point this feature will be redundant. Does this make sense?

I'll a deeper look into this in the coming days. Would be great if the final errors could be resolved, and then I'd try to merge this and the other PR in by say Tuesday evening 😊

andrewtavis:
Via the sync right now, @axif0 so you're aware, @catreedle and I discussed that we should remove autosuggestions from the normal get CLI command and have it be separate as the processes are fairly distinct. This will also be included in the PR :)

Thanks, @catreedle!

catreedle:
hi @andrewtavis @axif0 
I've added minimum coverage tests and removed autosuggestions from the interactive get prompt.
direct command like `scribe-data g -l swedish -dt autosuggestions` is still available.
I think this is ready for review. looking forward to your feedback. happy to revise if needed 😊

andrewtavis (approved):
All's looking great here, @catreedle! So wonderful to have your continued support of the project :) This really is a major feature 😊

I noticed that the profanity query isn't working, so I'll make an issue to investigate that :)

Issue #596, opened 2025-03-25
Upgrade Cmd Removes All Files Including Virtual Environment


axif0:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Running [upgrade.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/cli/upgrade.py) as,
```
scribe-data --upgrade
```
Removes all root files, including venv, instead of only updating necessary files which have changes in newer scribe version. Might be for this line, `shutil.rmtree(Path.cwd() / item.name)` 

Expected behavior might be 

- Only update required files.

- Preserve venv and other user-created files.

Also, add tests for [upgrade.py](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/cli/upgrade.py).

### Contribution
Happy to support this! I can also provide a review and check the PR once it's opened 😊

andrewtavis:
Feel free to add labels and the tags for the Scribe Board to new issues, @axif0 :)

meerhaasham1:
Hello @andrewtavis 
can I work on this

andrewtavis:
Sounds good, @meerhaasham1 :) Please let us know if you need any support!

meerhaasham1:
Hi @andrewtavis 
Just to update you, I have completed the task of identifying the issue with the `scribe-data --upgrade` command.

I’ve made the necessary changes to ensure that the command now behaves as expected, updating only the required files while preserving the `venv` and other user-created files.

Currently, I’m working on writing test cases for `test_upgrade.py` to ensure everything is functioning correctly.

I’ll update you once the tests are ready, and the pull request is submitted.

Thanks 

![Image](https://github.com/user-attachments/assets/c88f2284-4171-42b5-a1ab-c72eb2ab81e6)

andrewtavis:
Thanks for the update, @meerhaasham1! 😊

meerhaasham1:
Hi @andrewtavis,
Just a quick update — I’m currently working on the remaining issues related to “Add missing tests for files with low coverage” (#598).
[...]

PR #616, opened 2025-05-02
Fix file deletion and version parsing in upgrade.py

meerhaasham1:
### Contributor Checklist

- [x] Pull request is made from a [separate branch](https://docs.github.com/en/repositories/working-with-repositories/managing-branches-in-your-repository/creating-and-deleting-branches-within-your-repository), not the main branch  
- [x] Code has been tested using `pytest` as explained in the [contributing guide](https://github.com/scribe-org/Scribe-Data/blob/main/CONTRIBUTING.md#testing)  
- [x] I agree to follow Scribe-Data's Code of Conduct

---

### Description

**Improved File Safety During Upgrade**
- Prevents deletion of important folders like **`venv`**, **`.venv`**, and user files.
- Introduces **`MANAGED_DIRS`** and **`MANAGED_FILES`** to specify which files should be updated (e.g., `scribe_data/`, `setup.py`).
- Adds **`EXCLUDE_DIRS`** to skip folders like `.git` and virtual environments.
- Ensures only Scribe-managed files are updated while preserving everything else.

**Fixed Version Parsing to Prevent 404 Errors**
- Replaces version strings like `"Scribe-Data 4.1.0"` with just the number using **regex** (e.g., `4.1.0`).
- Builds the correct GitHub tag URL (e.g., **`v4.1.0`**) to avoid broken links during download.

---

### Testing

- Ran `scribe-data --upgrade` when current version is `4.1.0`  
  ✅ "You already have the latest version of Scribe-Data: 4.1.0"

- Mocked a newer version (`4.2.0`) → Confirmed:  
  ✅ `venv` and user files were preserved  
  ✅ Managed files were updated properly

- 🔧 Unit tests using `pytest` are pending, and I will add them in a separate PR.

---

### Related Issue

Fixes **#596**


andrewtavis (approved):
Hey @haasham1! 👋 Sorry for the wait on the review here :) Things have been quite busy. We made the decision to just do a basic call to pip for the upgrade instead of all of the work to download the tar ball and the like. Really appreciate your efforts here! Definitely was a catalyst for the eventual solution that we have now 😊

PR #626, opened 2025-08-04
Fix duplicate queries

### LINKED ISSUE #617: Query generation is including the same query form in more than one query
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

An issue that was discovered in https://github.com/scribe-org/Scribe-Data/pull/601 is that we have at times the same query form being generated in more than one query. An example of this is for English verbs:

- [english/verbs/query_verbs_1.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/wikidata/language_data_extraction/english/verbs/query_verbs_1.sparql)
- [english/verbs/query_verbs_4.sparql](https://github.com/scribe-org/Scribe-Data/blob/main/src/scribe_data/wikidata/language_data_extraction/english/verbs/query_verbs_4.sparql)

Each of the above query files is returning the `englishPastParticiple` form, which should not happen.

One potential fix for this is that when we find that there are new forms that haven't been included, it would be best if we regenerate all queries to make sure that no form is repeated. That way the list of available forms is directly translated to queries one to one 😊

harikrishnatp:
### Contributor checklist

---

### Description

This pull request resolves the issue of duplicate SPARQL query forms being generated. The fix involves update to the query generation script to ensure it correctly handles the current state of both local files and the live Wikidata dump.

Changes:

1. Made SPARQL parsing stateless : The script that reads existing .sparql files was modified to be stateless. It no longer accumulates data across runs, which was the original source of duplications
2. Corrected sub-language logic : While testing, macro languages were causing issues so query_processing logic in `check_missing_forms` was updated to correctly generate unique, filtered set of queries for each sub-language.
3. Improved error handling: No more crashing if it encounters a language in the Wikidata dump that is not present in the local project.

As part of this fix, all SPARQL query files were deleted and regenerated from the latest Wikidata lexeme dump. This not only fixed the duplication issue but also updated the queries to reflect the current state of data in wikidata. This led to the removal of some outdated queries (eg Ukrainian)  for which the data no longer appears to be available in the parsable format in the dump

**Testing:**
- The changes were tested by repeatedly running the generation script and verifying the output with a custom verification script.
- `pytest` was passing successfully. An outdated test (`test_list_data_types_specific_language`) was also updated to reflect the changes in the available data.

### Related issue

- Closes #617


andrewtavis:
@axif0, maybe we could look into this when we finalize the current PR for Scribe-Server as well? 😊

harikrishnatp:
Thanks so much for the encouragement, and apologies again for the delay in getting to this. Excited to see it merged! 😊

andrewtavis:
I think that there's value in both approaches - deleting the queries and keeping them. @axif0, what are the forms that are only available from the query service? Ideally we would be able to delete and regenerate, but then have those forms that are missing be accounted for in the process. This would assure that old fields that are from old labels or form combinations are no longer included in the queries :)

andrewtavis:
Closing this now given the conversation with @harikrishnatp and @axif0. Plan from here is that @harikrishnatp will be opening a new PR on this based on using the Wikidata Query Service to derive all form property combinations for language and data type pairs.

PR #629, opened 2025-08-17
test(check): added tests for check_project_metadata.py

GraceBocek:
### Contributor checklist

---

### Description

I added tests for check_project_metadata.py using the unittest framework, introducing tests/check/test_check_project_metadata.py. I tested by running the command `pytest`.

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/tests/check/test_check_project_metadata.py
+++ b/tests/check/test_check_project_metadata.py
@@ -1,0 +1,226 @@
+"""
+Tests functionality from check_project_metadata.py
+"""
+
+import unittest
+from io import StringIO
+from unittest.mock import patch
+
+from scribe_data.check.check_project_metadata import (
+    check_language_metadata,
+    get_available_languages,
+    get_missing_languages,
+    validate_language_properties,
+)
+
+
+class TestCheckProjectMetadata(unittest.TestCase):
+    def test_get_available_languages(self):
+        """
+        Tests that get_available_languages returns a dictionary with
+        languages from LANGUAGE_DATA_EXTRACTION_DIR
+        """
+        available_languages = get_available_languages()
+        desired_dict = {
+            "arabic": {},
+            "estonian": {},
+            "indonesian": {},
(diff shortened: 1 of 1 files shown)


GraceBocek:
@andrewtavis I believe I followed your instructions. Thank you for the feedback!

tests/check/test_check_project_metadata.py
```diff
@@ -0,0 +1,226 @@
...
+                }
+            },
+        }
+
+        # no missing languages
+        self.assertEqual(
+            get_missing_languages(reference_languages, target_languages), []
+        )
+
+        # more reference languages than target languages; should have no effect
```
andrewtavis, line 104: Minor comment here @GraceBocek, that for comments we try to have those on their own line be fully punctuated sentences including capitalization. So if it's its own line, then capitalize and include a period, and if it's inline then no capitalization and no period :) I'll get these this time!

tests/check/test_check_project_metadata.py
```diff
@@ -0,0 +1,228 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
```
andrewtavis, line 1: Added this to the PR as well, @GraceBocek :) For some licenses you're supposed to have a short version of the license at the top of the file, with GPL 3.0 being one of them as it's copy-left, so you need to make sure that people are aware that usage should be the same license in each file. There is a short version of the license that you can use, but it's also common practice these days to use the SPDX identifier as a one liner so that the files don't have large comments at the top with the license text.

andrewtavis: This fixed the [failed test here](https://github.com/scribe-org/Scribe-Data/actions/runs/17192462785/job/48769937752) :)

andrewtavis (approved):
Thanks so much for the great work on the tests here, @GraceBocek! Feel free to pick up another file/directory to write tests for :) As you heard in the call yesterday, we're still not there with the query generation files, but that should be done soon and then you could write tests for those as well 😊

Issue #641, opened 2025-10-10
Missing Sax Parsing Logic For Xml Files in extract_wiki.py file


priyankaforu:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

I am unable to parse the enwiki_dump files because of the parling logic that is missing , for sax parsing that we are doing for xml files 

I am expecting to complete the functions for triggering the call backs that captures the start and end elements / tags to extract the text

@axif0 @andrewtavis 

Please let me know if I can work on this issue and and move ahead with the scribe-data for auto suggestions 

![Image](https://github.com/user-attachments/assets/3c06a396-365e-486e-a0a4-33af498d815b)

PR #642, opened 2025-10-10
Fix/parsing logic error for sax

priyankaforu:
### Contributor checklist

---

### **Fix:** 

The enwiki's ".ndjson" was always parsing without any data, and this is because the parser is unable to open the file elements to parse like for example " `<page> <title> Heading For The Article </title> </page>` is not being read and called back to parse the array of articles/words that we need. So I have added the missing callback triggers for reading the texts inside those tags.

added filtering to omit the redirect links to various sources and name space pages

I have also verified it by downloading 2 enwiki_dump files and trying to parse them into words 👇

@axif0 @andrewtavis  @DeleMike Please have a look and lemme know if there are any corrections to be made or if my approach in understanding the issue is wrong

image

#641

### DIFF
--- a/src/scribe_data/wikipedia/extract_wiki.py
+++ b/src/scribe_data/wikipedia/extract_wiki.py
@@ -464,10 +464,6 @@ class WikiXmlHandler(xml.sax.handler.ContentHandler):
         self._current_tag = None
         self.target_articles = []
 
-    # Added the missing logic here ...
-    # Need startelement, endelement and characters to extract the text inside
-    # Sax must go through lines and trigger the callbacks so define them with start, end and characters
-
     def startElement(self, name, attrs):
         """
         Handle the start of an XML element.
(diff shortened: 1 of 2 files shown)


DeleMike:
I will review today. Well done! ✨

DeleMike (commented):
All looks okay @priyankaforu!

I would love to test it locally. I did not see the steps to reproduce the issue you had in #641, so it is hard for me to have the complete experience like you describe in the issue.

Could you add this, please?

priyankaforu:
> All looks okay @priyankaforu!
> 
> I would love to test it locally. I did not see the steps to reproduce the issue you had in #641, so it is hard for me to have the complete experience like you describe in the issue.
> 
> Could you add this, please?

Hey @DeleMike  , did you try downloading autosuggestions from wiki data ? Because for autosuggestions the issue is , parser is not parsing the characters inside the <page></page> tags, so to fix that we need to let Sax parser to read in between characters, and store them in the array. I just added the missing methods there :) If you still feel this is unclear, I can explain you better connecting with you, if possible whenever you have time

PR #652, opened 2025-11-04
Sparql query service approach

harikrishnatp:
### Contributor checklist

---

### Description

Sparql service implementation with the universal query being [https://w.wiki/FesH](https://w.wiki/FesH). and fallback query being [https://w.wiki/FesZ](https://w.wiki/FesZ). Also got rid of the duplicate query generation which was the main starting point of the issue. 

### Related issue

- #617

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,25 +1,14 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v6.0.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
       # - id: check-added-large-files
 
-  - repo: https://github.com/tcort/markdown-link-check
-    rev: v3.13.6
-    hooks:
-      - id: markdown-link-check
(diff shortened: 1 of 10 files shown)


harikrishnatp:
> Is it ok to skip `translations` data-type ? By querying `translations` script none of them works for now.

I thought about it too, some other data types are also like that. At first i didnt query them, just the main ones we used to query. But when i ran pytest, it showed that i needed to have those data types in the file structure. Thats why I decided to query them as well.

andrewtavis:
note: We can add `translations` into the data types in the list file to make sure that the test passes without needing to have a query for it :)

andrewtavis:
I'll take a look later tonight and bring it in, @harikrishnatp! Assuming as well that it's fine 😊

andrewtavis (approved):
praise: Really amazing work here, @harikrishnatp! Really is so great that we've come to this point with Scribe-Data! Can't wait to see how the queries look once we run the workflow 😊 Thanks as well for your support in reviewing, @axif0, and thank you both of the countless calls and syncs to get here! We'll see how it works, but from here we should be good to tweak the process as needed :)

PR #706, opened 2026-07-13
Parse multi-lang -wdp dumps in one pass and fix progress bars

axif0:
### Contributor checklist

---

### Description

- -wdp re-scanned the lexeme dump once per lang×type.
- Parse all languages/types in one pass instead. Drive progress from compressed bytes so bars always hit 100%.
- use same tqdm logic for wikidata and wiktionary

image

### DIFF
--- a/.github/workflows/ci_static_analysis.yaml
+++ b/.github/workflows/ci_static_analysis.yaml
@@ -24,7 +24,6 @@ on:
 
 jobs:
   format_check:
-    if: github.repository == 'scribe-org/Scribe-Data'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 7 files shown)


axif0:
@andrewtavis can you please help me with the complexity error issue? How can we fix it?

.github/workflows/ci_static_analysis.yaml
```diff
@@ -24,6 +24,7 @@ on:
 
 jobs:
   format_check:
+    if: github.repository == 'scribe-org/Scribe-Data'
```
andrewtavis: I think that it's fine if we run CI on other repos so people can know if their contributions that they have in their branch will fail CI? Has this been a problem for you, @axif0?

axif0: Oh, we can revert it then.

andrewtavis (approved):
praise: All's looking great to me, @axif0! Thanks for all the care you're putting into the project recently :) Really is great to have these PRs coming in 😊

Issue #713, opened 2026-09-08
Add Data Contracts flow (export_contracts, check_contracts, filter_data) to ARCHITECTURE.md


haasham1:
### Terms

- [x] I have searched all [open documentation issues](https://github.com/scribe-org/Scribe-Data/issues?q=is%3Aopen+is%3Aissue+label%3Adocumentation)
- [x] I agree to follow Scribe-Data's [Code of Conduct](https://github.com/scribe-org/Scribe-Data/blob/main/.github/CODE_OF_CONDUCT.md)

### Current Documentation

ARCHITECTURE.md documents the download, get, convert, total, and list command flows, but the Data Contracts feature isn't represented anywhere in the diagram there are no nodes or arrows for it, even though export_contracts, check_contracts, and filter_data are real, working commands documented as a top-level section in the README.

### Suggested Enhancement

Add the Data Contracts flow to the architecture diagram:

A per-language contract file (e.g. de.yaml) is pulled out via export_contracts
That contract is then used one of two ways:
 - check_contracts verifies that a get export is complete against the contract
 - filter_data trims an export down to only what the contract calls for

Note: once this is resolved here, the main cross-repo architecture diagram will also need this same Data Contracts flow added, so we don't lose track of it.

@andrewtavis

andrewtavis:
Closed by #714

PR #714, opened 2026-09-08
[Feat] Updates in Architecture.md 

haasham1:
**Closes** [#713](https://github.com/scribe-org/Scribe-Data/issues/713)

Adds the missing Data Contracts flow to the **ARCHITECTURE.md** Mermaid diagram. Previously, **export_contracts,** **check_contracts**, and **filter_data** were real, documented commands with no representation in the architecture diagram.

Added a new **CONTRACTS_FLOW** subgraph showing:
A per-language contract file as the source node
**export_contracts** pulling the contract out
**check_contracts** verifying a GET export against the contract
**filter_data** trimming a GET export down to only what the contract calls for
Connected the new subgraph to the existing JSON files output node, since contracts operate on get exports
Kept node/edge naming and label conventions consistent with the rest of the diagram


andrewtavis (approved):
I removed the check contracts command from the diagram as I felt like it was getting to complex, @haasham1. Also made a few edits and updated the changelog to reflect the addition of the whole architecture diagram. Thanks for this!

